PR #4, opened 2020-01-31
Introduces Graph QL

KaiVolland:
This introduces `Graph QL` and the `Graphi QL` UI.

graphiql


shogun-lib/pom.xml
```diff
@@ -100,6 +100,28 @@
       <artifactId>hibernate-types-52</artifactId>
     </dependency>
 
+    <!-- GraphQL -->
+    <dependency>
+      <groupId>com.graphql-java</groupId>
+      <artifactId>graphql-spring-boot-starter</artifactId>
+      <version>5.0.2</version>
```
buehner: all versions are set in the dependencyManagement section of the parent pom. in the submodules pom we do not have to define versions. so please migrate the version info

KaiVolland: affirmative

PR #11, opened 2020-03-02
Adds ConditionalOnMissingBean annotation

KaiVolland:
This adds the `ConditionalOnProperty` annotation to the GraphQL Beans.

To skip the creation of the SHOGun GraphQL bean you have to set a property (in `application.properties`) like this:

```
shogun.graphql.skipBean=true
```

The project representations for the GraphQL beans should then have an annotation like this:
```java
    @Bean
    @ConditionalOnProperty(
        value="shogun.graphql.skipBean",
        havingValue = "true"
    )
    @Override
    public GraphQL graphQL() {
        return graphQL;
    }
```
I'd really love to have this code smarter so you don't have to set a property when extending the SHOGun graphql, but i didn't found a way.
Would be really nice if some can double check this.


KaiVolland:
> Nice solution!

Sadly it doesn't seem to work. I can reconstruct it in another project.

PR #12, opened 2020-03-04
Introduces Geometry scalar for GraphQL

KaiVolland:
This introduces the [custom scalar type](https://www.graphql-java.com/documentation/v12/scalars/) `Geometry`. This way we can use `Geometry` in the GraphQL schema instead of `String`.
So geometries can now be fetched as GeoJSON instead of WKT.

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/GeometryScalar.java
+++ b/shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/GeometryScalar.java
@@ -26,7 +26,7 @@ public class GeometryScalar {
 
         @Override
         public Object parseValue(Object input) {
-            return parseGeoemtryFromVariable(input);
+            return parseGeometryFromVariable(input);
         }
 
         @Override
@@ -47,14 +47,14 @@ public class GeometryScalar {
             } catch (JsonProcessingException e) {
                 LOG.error("JSON Processing error while writing geometry for GraphQL");
                 LOG.trace("Full stack trace: ", e);
+                throw new CoercingSerializeException(e.getMessage());
             }
         } else {
             throw new CoercingSerializeException("Unable to serialize " + dataFetcherResult + " as a geometry");
         }
-        return null;
     }
 
-    private static Object parseGeoemtryFromVariable(Object input) {
+    private static Object parseGeometryFromVariable(Object input) {
         if (input instanceof String) {
             String geometryString = (String)input;
             try {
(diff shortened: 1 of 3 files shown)


shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/GeometryScalar.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private static boolean isAGeometry(Object possibleGeometryValue) {
+        return possibleGeometryValue instanceof Geometry;
+    }
+
+    private static Object serializeGeometry(Object dataFetcherResult) {
+        if (isAGeometry(dataFetcherResult)) {
+            Geometry geometry = (Geometry) dataFetcherResult;
+            ObjectMapper mapper = new ObjectMapper();
+            mapper.registerModule(new JtsModule());
```
dnlkoch: Please check if we can reuse the global `objectMapper` here.

shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/GeometryScalar.java
```diff
@@ -0,0 +1,80 @@
...
+            try {
+                return mapper.readValue(mapper.writeValueAsString(geometry), HashMap.class);
+            } catch (JsonProcessingException e) {
+                e.printStackTrace();
```
dnlkoch: Please make use of the `Logger` here or let the method throw a `JsonProcessingException`.

shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/GeometryScalar.java
```diff
@@ -0,0 +1,84 @@
...
+                LOG.error("JSON Processing error while writing geometry for GraphQL");
+                LOG.trace("Full stack trace: ", e);
+            }
+        } else {
+            throw new CoercingSerializeException("Unable to serialize " + dataFetcherResult + " as a geometry");
+        }
+        return null;
+    }
+
+    private static Object parseGeoemtryFromVariable(Object input) {
```
hwbllmnn:
```suggestion
    private static Object parseGeometryFromVariable(Object input) {
```

shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/GeometryScalar.java
```diff
@@ -0,0 +1,84 @@
...
+        return possibleGeometryValue instanceof Geometry;
+    }
+
+    private static Object serializeGeometry(Object dataFetcherResult) {
+        if (isAGeometry(dataFetcherResult)) {
+            Geometry geometry = (Geometry) dataFetcherResult;
+            try {
+                return objectMapper.readValue(objectMapper.writeValueAsString(geometry), HashMap.class);
+            } catch (JsonProcessingException e) {
+                LOG.error("JSON Processing error while writing geometry for GraphQL");
```
ahennr: Maybe you could throw `e` as a `CoercingSerializeException` here as well as in the other overrides?

PR #15, opened 2020-03-10
Adds missing config and schema fields

KaiVolland:
This adds needed `spring.jpa.database-platform` see: https://stackoverflow.com/a/59623399/10342669

It also adds some missing fields to the GraphQL schema.


shogun-lib/src/main/resources/graphql/shogun.graphqls
```diff
@@ -12,6 +12,8 @@ type Query {
 
 type Application {
     id: ID
+    created: String
```
ahennr: `Date` is more appropriate here, isn't it?

shogun-lib/src/main/resources/graphql/shogun.graphqls
```diff
@@ -32,9 +36,18 @@ type Layer {
...
     details: JSON
     clientConfig: JSON
 }
+
+type UserGroup {
+    id: ID
+    created: String
+    modified: String
+    name: String
+}
```
ahennr: minor: plz add newline.

shogun-lib/src/main/resources/graphql/shogun.graphqls
KaiVolland: I had issues with configuring this as Date. And as a Date ist transformed to a String for the JSON response anyway this should be ok for now.

ahennr: Alright.

PR #18, opened 2020-03-12
Updates FlyWay SQLs according to KeyCloak

KaiVolland:
This updates the flyway sqls according to the changes to the models of `User`, `Role`, etc.

It includes a hack to the SecurityContextUtil.

Currently the `userRepository.findByKeycloakId` returns an user but it should return an `Optional`. We should fix this is a followup PR.


shogun-boot/src/main/resources/db/migration/V1.0.2__Init_admin.sql
```diff
@@ -0,0 +1,60 @@
...
+--);
+--
+--INSERT INTO userclasspermissions VALUES (
+--    nextval('hibernate_sequence'),
+--    NOW()::timestamp,
+--    NOW()::timestamp,
+--    'de.terrestris.shogun.lib.model.Group',
+--    (SELECT id FROM permissions WHERE name = 'ADMIN'),
+--    (SELECT id FROM users WHERE username = 'admin')
+--);
```
dnlkoch: Is there any reason why we keep this file?

KaiVolland: Nope. I'll remove it.

shogun-boot/src/main/resources/db/migration/V1.0.0__Baseline.sql
```diff
@@ -53,41 +53,24 @@ CREATE TABLE files (
...
     details jsonb,
-    email text NOT NULL,
-    enabled boolean,
-    password text,
-    username text NOT NULL,
     CONSTRAINT users_pkey PRIMARY KEY (id),
-    CONSTRAINT users_unique_email UNIQUE (email),
-    CONSTRAINT users_unique_id UNIQUE (id),
-    CONSTRAINT users_unique_username UNIQUE (username)
+    CONSTRAINT users_unique_id UNIQUE (id)
```
hwbllmnn: You can simplify the unique and pk constraints by using `id bigserial` and annotating the column with
```
@Column
@GeneratedValue(strategy = GenerationType.IDENTITY)
@Id
```

shogun-boot/src/main/resources/db/migration/V1.0.0__Baseline.sql
```diff
@@ -53,41 +53,24 @@ CREATE TABLE files (
...
 );
 
 CREATE TABLE groups (
     id bigint NOT NULL,
```
hwbllmnn: Same here

shogun-boot/src/main/resources/db/migration/V1.0.0__Baseline.sql
ahennr: Herewith, sequential IDs would be generated automatically per table, wouldn't they?

KaiVolland: so i can remove the constraint lines?

hwbllmnn: Yep. `id bigserial` (64 bit int) or `id serial` (32 bit int) will add the constraints, set the field as primary key and add a new sequence. This has the added benefit that objects in all tables start with id `1`.

KaiVolland: I'll change this in an follow up PR as this affects all ids.

buehner:
We will address the hints of @hwbllmnn in a follow up PR

PR #20, opened 2020-03-13
Specifications removal for "simple" attributes

ahennr:
Use spring internal JPA mapping for single attributes and return optional.

Plz review.


shogun-lib/src/main/java/de/terrestris/shogun/lib/listener/LoginListener.java
```diff
@@ -53,8 +54,8 @@ public void onApplicationEvent(InteractiveAuthenticationSuccessEvent event) {
 
         // add missing groups to shogun db
         userGroups.stream().map(GroupRepresentation::getId).forEach(keycloakGroupId -> {
-            Group group = groupRepository.findByKeycloakId(keycloakGroupId);
-            if (group == null) {
+            Optional<Group> group = groupRepository.findByKeycloakId(keycloakGroupId);
+            if (! group.isPresent()) {
```
hwbllmnn:
```suggestion
            if (!group.isPresent()) {
```

PR #28, opened 2020-03-25
Check for subject (user id) in token object if ID token is null

ahennr:
Title says it all.

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/listener/LoginListener.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/listener/LoginListener.java
@@ -6,6 +6,9 @@ import de.terrestris.shogun.lib.repository.GroupRepository;
 import de.terrestris.shogun.lib.repository.UserRepository;
 import de.terrestris.shogun.lib.security.SecurityContextUtil;
 import org.keycloak.KeycloakPrincipal;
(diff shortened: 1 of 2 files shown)


shogun-lib/src/main/java/de/terrestris/shogun/lib/listener/LoginListener.java
```diff
@@ -40,7 +43,15 @@ public void onApplicationEvent(InteractiveAuthenticationSuccessEvent event) {
...
-        String keycloakUserId = keycloakPrincipal.getKeycloakSecurityContext().getIdToken().getSubject();
+        KeycloakSecurityContext keycloakSecurityContext = keycloakPrincipal.getKeycloakSecurityContext();
+        IDToken idToken = keycloakSecurityContext.getIdToken();
+        String keycloakUserId;
+        if (idToken != null) {
+            keycloakUserId = idToken.getSubject();
+        } else {
+            AccessToken accessToken = keycloakSecurityContext.getToken();
+            keycloakUserId = accessToken.getSubject();
+        }
```
dnlkoch: This looks *very* similiar to the changes/code in the `SecurityContextUtil`. Do you see any chance to reuse the code?

ahennr: Alright, see 57e97d8

PR #31, opened 2020-03-27
Introduces findAllBy in BaseService

KaiVolland:
~~This introduces the possibility to call the `findAll` method with a `Specification`.~~

This introduces the `findAllBy` method which can be called with a `Specification`.

This may replace the currently outcommented functions `findBy` and `findByFilter`. If the majority of the devs aggree id remove these commented lines.

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
@@ -29,42 +29,6 @@ public abstract class BaseService<T extends BaseCrudRepository<S, Long> & JpaSpe
     @Autowired
     ObjectMapper objectMapper;
 
-//    public List<S> findByFilter(String column, String filter) {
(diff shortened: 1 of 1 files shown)


dnlkoch:
Generally approved, but sementically I would prefer to have an additional signature like 
```
getAllBy(Specification specification)
```

KaiVolland:
> Generally approved, but sementically I would prefer to have an additional signature like
> 
> ```
> getAllBy(Specification specification)
> ```

Any specific reason why? To enhance readability and make code clearer? Or is there anything bad about that polymorphic stuff?

dnlkoch:
Yes, just because for better readability, but it's not a blocker and no, there is nothing bad about the polymorphism. You decide.

KaiVolland:
Already refactored it. I was just curious. :)

KaiVolland:
> LGTM, however I would prefer to user interfaces in repositories wherever it is possible.

Our repositories extend the `JpaSpecificationExecutor` interface which provides this functionality.

PR #39, opened 2020-04-07
Introduces KeycloakUtil, modifies ApplicationInfo and readds UserController

KaiVolland:
This introduces the `KeycloakUtil`, modifies the `ApplicationInfo` and readds the `UserController`.

Applicationinfo now contains the userId instead of the userName.

This also fixes the locale messages.


shogun-lib/src/main/java/de/terrestris/shogun/lib/security/SecurityContextUtil.java
```diff
@@ -78,17 +76,25 @@
...
+
+            if (idToken != null) {
+                keycloakUserId = idToken.getSubject();
+            } else {
+                AccessToken accessToken = keycloakSecurityContext.getToken();
+                keycloakUserId = accessToken.getSubject();
+            }
+
+            return keycloakUserId;
+        } else if (authentication.getPrincipal() instanceof String) {
```
hwbllmnn, line 93: Since you also return `null` in the generic else branch you can omit this

PR #40, opened 2020-04-09
Introduce GeoServer importer module

dnlkoch:
This adds a new submodule `shogun-gs-importer` for managing uploads to the GeoServer via the `importer` extension.

Basically this is a port of the existing code base found in [shogun-core](https://github.com/terrestris/shogun-core/tree/master/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/importer).


dnlkoch:
Since there is no progress (and obviously no need for it) I'll close this one.

PR #42, opened 2020-04-16
OgcXMLutil: Override inputstream

ahennr:
This PR introduces a helper function overriding the `InputStream` of a request with given XML document.

plz review @terrestris/devs

### DIFF
--- a/shogun-gs-interceptor/src/main/java/de/terrestris/shogun/interceptor/util/OgcXmlUtil.java
+++ b/shogun-gs-interceptor/src/main/java/de/terrestris/shogun/interceptor/util/OgcXmlUtil.java
@@ -1,6 +1,7 @@
 package de.terrestris.shogun.interceptor.util;
 
 import de.terrestris.shogun.interceptor.exception.InterceptorException;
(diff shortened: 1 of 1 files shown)


shogun-gs-interceptor/src/main/java/de/terrestris/shogun/interceptor/util/OgcXmlUtil.java
```diff
@@ -161,4 +163,29 @@ public static Element getDocumentElement(Document doc) {
...
+        ) {
+            Result outputTarget = new StreamResult(outputStream);
+            TransformerFactory.newInstance().newTransformer().transform(xmlSource, outputTarget);
+
+            try (InputStream is = new ByteArrayInputStream(outputStream.toByteArray())) {
+                request.setInputStream(is);
+            }
+            return request;
+        } catch (TransformerException | IOException e) {
+            LOG.error("Error on trying to parse an xml body.", e);
```
hwbllmnn, line 186:
```suggestion
            LOG.error("Error on trying to parse an xml body.");
            LOG.trace("Stack trace:", e);
```

PR #43, opened 2020-04-16
Ignore CSRF geoserver.action of interceptor

ahennr:
Title says it all.

Plz review @terrestris/devs

### DIFF
--- a/shogun-gs-interceptor/src/main/java/de/terrestris/shogun/interceptor/config/InterceptorWebSecurityConfig.java
+++ b/shogun-gs-interceptor/src/main/java/de/terrestris/shogun/interceptor/config/InterceptorWebSecurityConfig.java
@@ -3,6 +3,7 @@ package de.terrestris.shogun.interceptor.config;
 import de.terrestris.shogun.config.WebSecurityConfig;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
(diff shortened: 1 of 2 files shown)


shogun-gs-interceptor/src/main/java/de/terrestris/shogun/interceptor/config/InterceptorWebSecurityConfig.java
```diff
@@ -19,10 +20,13 @@ protected void customHttpConfiguration(HttpSecurity http) throws Exception {
...
-            .permitAll()
+                .permitAll()
             .antMatchers("/interceptorrules/**")
                 .hasRole("interceptor-admin")
             .anyRequest()
-                .authenticated();
+                .authenticated()
+            .and()
+                .csrf()
+                .ignoringAntMatchers("/geoserver.action/**");
```
dnlkoch, line 30: And why do we want this?

dnlkoch: If we want to disable CSRF protection for specfic use cases (e.g. requests via the swagger ui or `curl`), we may want to use a specific matcher (see e.g. [here](https://github.com/terrestris/shogun/pull/40/files#diff-cd40bf5efa26b6504a955a6ad418d59fR14)).

ahennr:
Will move this to project specific interceptor config.

PR #44, opened 2020-04-20
Add userRepresentation to getUserBySession

LukasLohoff:
Sets `userRepresentation` for the user returned by `getUserBySession`. Based on the userService methods which also return a `userRepresentation`.

@devs Please review


LukasLohoff:
Thanks!

PR #46, opened 2020-05-29
Fix column definition for UUID type

dnlkoch:
Set column definition to ensure correct creation of the column via Hibernate. This won't have any effect to existing db installations, it'll just ensure a valid creation of the database via Hibernate (if ever needed…).


KaiVolland:
:+1:

PR #49, opened 2020-06-02
Update baseline (!)

dnlkoch:
This updates the baseline to create the tables if needed only.

Yes, changing an existing baseline is against the best-practices, but I think the current baseline might also produce conflicts in the future. 

At the current state existing installations should fix the update while removing the appropriate entry from the `flyway_schema_history` table.

@terrestris/devs What do you think?


dnlkoch:
Updated the generation of primary keys and constraints, but as suggested [here](https://vladmihalcea.com/postgresql-serial-column-hibernate-identity/) I would propose to leave the ID generation as is (via the hibernate sequence).

dnlkoch:
Final call: Any projects that might get into trouble after applying this diff @terrestris/devs?

PR #50, opened 2020-06-03
WIP Better handling of File and ImageFile entities

dnlkoch:
:construction_worker:


shogun-config/src/main/resources/application-base.yml
```diff
@@ -56,3 +56,23 @@ keycloakauth:
   password: shogun
   master-realm: master
   admin-client: admin-cli
+
+upload:
+  file:
+    supportedContentTypes:
```
KaiVolland, line 62: Can this be set to `all` or `*` in a childproject? Otherwise i would like to remove this restriction in shogun.

dnlkoch: Not yet, but I think this a very good idea. I'l check that.

PR #51, opened 2020-06-09
WIP Initialize group REST controller

dnlkoch:
This initializes a REST controller for the `Group` entity and also adds some special endpoints to get the associated Keycloak informations:


shogun-lib/src/main/java/de/terrestris/shogun/lib/listener/LoginListener.java
```diff
@@ -47,21 +58,28 @@ public synchronized void onApplicationEvent(InteractiveAuthenticationSuccessEven
...
-        // add missing user to shogun db
+        // Add missing user to shogun db
         Optional<User> userOptional = userRepository.findByKeycloakId(keycloakUserId);
         User user = userOptional.orElse(null);
         if (user == null) {
             user = new User(keycloakUserId, null, null, null);
             userRepository.save(user);
+
+            // Add admin instance permissions for the user.
+            userInstancePermissionService.setPermission(user, user, PermissionCollectionType.ADMIN);
```
KaiVolland, line 69: Looks like every user becomes an admin but probably i just don't get what happens here...

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/GroupService.java
```diff
@@ -0,0 +1,119 @@
...
+            if (user.isPresent()) {
+                user.get().setKeycloakRepresentation(groupMember);
+                users.add(user.get());
+            }
+        }
+
+        return users;
+    }
+
+    private Group setTransientKeycloakRepresentations(Group group) {
```
KaiVolland, line 110: Is this some kind of semi-automated update process? I'm not sure if i like that this is called on every service method. Is this really necessary?

shogun-lib/src/main/java/de/terrestris/shogun/lib/listener/LoginListener.java
marcjansen: Please double check!

weskamm: As stated above `Newly created users will get the Permssion.ADMIN on their own instances.` So the user is able to administer himself (update, delete)

dnlkoch: No, actually this saves the `ADMIN` permission to the instance of the user itself - it is not connected to any other user or role. Without this line, the user wouldn't be allowed to execute any instance related read, update or delete operation.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/GroupService.java
dnlkoch: This only sets the transient field (the group details saved in Keycloak) in the group instance.

PR #53, opened 2020-06-25
Enable revisioning of entities

dnlkoch:
This enables auditing entity changes via [envers](https://hibernate.org/orm/envers/) / [spring-data-envers](https://spring.io/projects/spring-data-envers) for all SHOGun entities (except the `File` and `ImageFile` ones).

For every change a snapshot of the whole entity will be stored in separate table inside a new schema `shogun_rev`. The following screenshot demonstrates this for the `Application` after the name has been changed from `string` to `name`:

Bildschirmfoto vom 2020-06-25 13-53-30

To get the revisions or a specific revision for a given entity the following endpoints have been added (for each entity):

* `/entity/{id}/rev`
  * Returns all revisions for the given entity.
* `/entity/{id}/rev/{rev}`
  * Returns the given revision for the given entity.
* `/entity/{id}/lastrev`
  * Returns the latest revision for the given entity.

The following screenshot shows the newly added endpoints in the Swagger UI:

Bildschirmfoto vom 2020-06-25 13-48-43

Whereas the example response would look like:

```
{
  "metadata": {
    "revisionDate": "2020-06-25T13:51:10.864",
    "revisionNumber": 22,
    "revisionInstant": "2020-06-25T11:51:10.864Z",
    "delegate": {
      "id": 22,
      "timestamp": 1593085870864,
      "revisionDate": "2020-06-25T11:51:10.864+0000"
    },
    "revisionType": "UNKNOWN",
    "requiredRevisionNumber": 22,
    "requiredRevisionInstant": "2020-06-25T11:51:10.864Z"
  },
  "entity": {
    "id": 19,
    "created": "2020-06-25T11:50:42.987+0000",
    "modified": "2020-06-25T11:51:10.848+0000",
    "name": "name",
    "i18n": {
      "de": "string"
    },
    "stateOnly": true,
    "clientConfig": {
      "logoPath": "string",
      "mapView": {}
    },
    "layerTree": {},
    "layerConfig": {},
    "toolConfig": {}
  },
  "revisionNumber": 22,
  "requiredRevisionNumber": 22,
  "revisionInstant": "2020-06-25T11:51:10.864Z",
  "requiredRevisionInstant": "2020-06-25T11:51:10.864Z"
}
```


marcjansen:
Very interesting! Looking fine on first sight. Others should chime in.

shogun-boot/src/main/java/de/terrestris/shogun/boot/runner/ApplicationInitializer.java
```diff
@@ -0,0 +1,39 @@
...
+     * should be imported through the existing flyway migrations.
+     */
+    private void checkExistenceOfPermissions() {
+        Iterable<PermissionCollection> permissions = permissionCollectionRepository.findAll();
+
+        long permissionsLength = StreamSupport.stream(permissions.spliterator(), false).count();
+
+        if (permissionsLength == 0) {
+            log.fatal("Couldn't find any permissions in the database. This will most likely lead to errors. Please " +
+                "ensure flyway is enabled your application.yaml! Alternatively the permissions may be inserted " +
```
marcjansen, line 35:
```suggestion
                "ensure flyway is enabled in your application.yaml! Alternatively the permissions may be inserted " +
```

shogun-config/src/main/resources/application-base.yml
```diff
@@ -13,12 +13,25 @@ spring:
   main:
     banner-mode: off
   jpa:
+    open-in-view: true
```
marcjansen, line 16: I faintly remember this was some source of a performance penalty in the stone ages… I suspect we actually need thsi, correct?

marcjansen: Please see [this article](https://www.baeldung.com/spring-open-session-in-view), and then choose an option.

dnlkoch: Thanks for the note and I think we should discuss the usage of the session in view in general. Actually the change in here is just the default for spring boot apps and hides a message on application startup.

dnlkoch: And yes, currently we need this setting. Otherwise the lazy initialized references must be fetched manually.

dnlkoch:
Thanks for the reviews!

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/security/permission/GroupClassPermission.java
```diff
@@ -2,13 +2,17 @@
...
+import org.hibernate.envers.Audited;
 
 import javax.persistence.Entity;
 import javax.persistence.ManyToOne;
 import javax.persistence.Table;
 
 @Entity(name = "groupclasspermissions")
 @Table(schema = "shogun")
+@Audited
+@AuditTable(value = "groupclasspermissions_rev", schema = "shogun_rev")
```
weskamm, line 15: do we really need revisions on permissions? i am not sure..

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
```diff
@@ -56,6 +58,24 @@
         return repository.findById(id);
     }
 
+    @PreAuthorize("hasRole('ROLE_ADMIN') or hasPermission(#entity, 'READ')")
```
weskamm, line 61: if the construct still exists, i would like to use something more generic:
`@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName())")`

weskamm:
just minor comments and now its merged already, so its ok :-)

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
dnlkoch: Currently the `@configHolder.getSuperAdminRoleName()` doesn't exist (yet), but I agree that we should reimplement it! Good catch.

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/security/permission/GroupClassPermission.java
dnlkoch: Yes, tricky question. Right now the permissions will be instantiated/deleted on entity creation/deletion and it's not even possible to change the permissions directly through any web interface. So - more specifically - the audit table is actually of no use. 
But: I do have a feature branch in progress that implements the appropriate web interfaces to update the permissions for each entity and so the tables *might* become more useful in the near future.

weskamm: ok we will see if there is any practical use for this. nice work!

PR #56, opened 2020-08-05
Init jackson ID resolver

dnlkoch:
This adds the `BaseEntityIdResolver` class that can be used to implement custom jackson ID serializers (see the `ImageFileIdResolver` class as an example).


dnlkoch:
Great, thanks for the review!

PR #54, opened 2020-06-30
Add endpoints to read, create, update and delete entity permissions

dnlkoch:
This initializes the `BasePermissionController` which adds all relevant CRUD endpoints to manage all kind permissions (`UserInstancePermission`, `UserClassPermission`, `GroupInstancePermission`, `GroupClassPermission`) on entity level:

#### READ

* `@GetMapping("/{id}/permissions/class/group/{groupId}")`
  * Returns the `GroupClassPermission` for the given `Group` and entity.
* `@GetMapping("/{id}/permissions/class/user/{userId}")`
  * Returns the `UserClassPermission` for the given `User` and entity.
* `@GetMapping("/{id}/permissions/instance/group/{groupId}")`
  * Returns the `GroupInstancePermission` for the given `Group` and entity.
* `@GetMapping("/{id}/permissions/instance/user/{userId}")`
  * Returns the `UserInstancePermission` for the given `User` and entity.
* `@GetMapping("/{id}/permissions/class/group")`
  * Returns all `GroupClassPermission` for the given entity.
* `@GetMapping("/{id}/permissions/class/user")`
  * Returns all `UserClassPermission` for the given entity.
* `@GetMapping("/{id}/permissions/instance/group")`
  * Returns all `GroupInstancePermission` for the given entity.
* `@GetMapping("/{id}/permissions/instance/user")`
  * Returns all `UserInstancePermission` for the given entity.

#### CREATE/UPDATE

* `@PostMapping("/{id}/permissions/class/group/{groupId}")`
  * Creates/Updates the `GroupClassPermission` for the given `Group` and entity (via the provided `PermissionType` in the request body).
* `@PostMapping("/{id}/permissions/class/user/{userId}")`
  * Creates/Updates the `UserClassPermission` for the given `Group` and entity (via the provided `PermissionType` in the request body).
* `@PostMapping("/{id}/permissions/instance/group/{groupId}")`
  * Creates/Updates the `GroupInstancePermission` for the given `Group` and entity (via the provided `PermissionType` in the request body).
* `@PostMapping("/{id}/permissions/instance/user/{userId}")`
  * Creates/Updates the `UserInstancePermission` for the given `Group` and entity (via the provided `PermissionType` in the request body).

#### DELETE

* `@DeleteMapping("/{id}/permissions/class/group/{groupId}")`
  * Deletes the `GroupClassPermission` for the given `Group` and entity.
* `@DeleteMapping("/{id}/permissions/class/user/{userId}")`
[...]

PR #57, opened 2020-08-11
adjust logging level for deleted permissions

LukasLohoff:
This tones down the logging when deleting permissions.

I saw `getGenericClassName` being used in other classes to log the entity type. Can we include that here as well?

@devs Please review


LukasLohoff:
Just saw that this is already included in #54 
So I'm closing this

dnlkoch:
As #54 is a proof of concept (and the logger is annoying you badly), please reopen and merge.

LukasLohoff:
Alright, done!

PR #59, opened 2020-08-26
Enable Spring Boot actuator endpoints

dnlkoch:
This enables all endpoints from the [Spring Boot Actuator](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-endpoints) (except the `shutdown` one) and adds some notes about its usage to the README.

### DIFF
--- a/shogun-boot/src/main/java/de/terrestris/shogun/boot/config/BootWebSecurityConfig.java
+++ b/shogun-boot/src/main/java/de/terrestris/shogun/boot/config/BootWebSecurityConfig.java
@@ -58,7 +58,8 @@ public class BootWebSecurityConfig extends WebSecurityConfig {
                 .csrf()
                     .csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse())
                     .ignoringRequestMatchers(csrfRequestMatcher)
-                    .ignoringAntMatchers("/graphql");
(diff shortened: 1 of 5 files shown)


marcjansen (approved):
Awesome. I am +1 on merging!

shogun-boot/src/main/java/de/terrestris/shogun/boot/config/BootWebSecurityConfig.java
```diff
@@ -35,6 +35,11 @@ protected void customHttpConfiguration(HttpSecurity http) throws Exception {
                     "/index.html"
                 )
                     .permitAll()
+                .antMatchers(
+                    "/swagger-ui.html",
```
hwbllmnn, line 39: The `swagger-ui.html` could IMO also be accessible for all, the actual requests it can send would still be restricted anyway.

hwbllmnn (approved):
Nice, just one suggestion

shogun-boot/src/main/java/de/terrestris/shogun/boot/config/BootWebSecurityConfig.java
dnlkoch: Agreed, fixed in 8c14b0a.

dnlkoch:
Thanks for the reviews!

PR #97, opened 2020-08-27
Fix spring security versions

dnlkoch:
As artifact `spring-context-support` belongs to the `org.springframework`, the spring version should be used instead of the spring-security version.

This also updates the spring-security version.

**NOTE**: Due to the update it _might_ become necessary to invalidate all sessions in projects to fix version incompatibility errors (watch out for `java.io.InvalidClassException` exceptions)!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -316,6 +316,11 @@
         <version>${spring.version}</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


buehner:
:+1: LGTM

hwbllmnn (approved):

PR #125, opened 2020-10-05
Enable second level and query caching via ehcache

dnlkoch:
This suggest to enable the second level and query cache for all entities via ehcache.

In addition a new interface `/cache/evict` is introduced to clear the current cache.

### DIFF
--- a/shogun-config/src/main/resources/application-base.yml
+++ b/shogun-config/src/main/resources/application-base.yml
@@ -13,6 +13,8 @@ spring:
   main:
     banner-mode: off
   jpa:
+    # Show SQL statements
+    show-sql: true
     open-in-view: true
     database-platform: org.hibernate.spatial.dialect.postgis.PostgisPG95Dialect
     hibernate:
@@ -21,6 +23,11 @@ spring:
           non_contextual_creation: true
       ddl-auto: none
     properties:
+      javax:
+        persistence:
+          sharedCache:
+            # Only entities with @Cacheable annotation will use L2 cache
+            mode: ENABLE_SELECTIVE
       org:
         hibernate:
           # See https://docs.jboss.org/envers/docs/#configuration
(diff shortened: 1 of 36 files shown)


shogun-config/src/main/resources/application-base.yml
```diff
@@ -13,6 +13,8 @@ spring:
   main:
     banner-mode: off
   jpa:
+    # Show SQL statements
+    show-sql: true
```
weskamm, line 17: intended?

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/Topic.java
```diff
@@ -43,6 +55,7 @@
     @OneToOne
     @Fetch(FetchMode.JOIN)
     @Audited(targetAuditMode = RelationTargetAuditMode.NOT_AUDITED)
+    @Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
```
weskamm, line 58: not sure why used explicitly on a single field here while enabled globally on top

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/Topic.java
```diff
@@ -59,6 +72,7 @@
     @Type(type = "jsonb")
     @Column(columnDefinition = "jsonb", name = "searchlayerconfigs")
     @Basic(fetch = FetchType.LAZY)
+    @Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
```
weskamm, line 75: as above

weskamm:
:rocket: 
just 2 minor questions

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/Topic.java
dnlkoch: This is because collections are not cached by default and need to be explicitly marked as cacheable.

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/Topic.java
dnlkoch: See above.

shogun-config/src/main/resources/application-base.yml
dnlkoch: Actually it is as this will show us:

  * if hibernate creates (unwanted) complex queries
  * if a query is being issued to the database even if it is not intended

during development.

dnlkoch:
Thanks for the review!

PR #139, opened 2020-10-12
Add paging and move permissions check to database

LukasLohoff:


LukasLohoff:
Closed in favor of #606

PR #146, opened 2020-10-22
Enhance configuration of jib - plugin for docker-builds

ahennr:
In particular:

* skip jib for libs / configs and declare main classes explicitly
* use digest / SHA 256 as image reference to get always the same image for each build

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -217,7 +217,7 @@
           </executions>
           <configuration>
             <from>
-              <image>openjdk:11-slim</image>
+              <image>openjdk@sha256:2b6eb2c178908b6704d82258652806eca3c0cd9261df7f5edcc3486d63bc6e60</image>
             </from>
             <to>
               <image>nexus.terrestris.de/shogun/${project.artifactId}:latest</image>
--- a/pom.xml
+++ b/pom.xml
@@ -223,6 +223,8 @@
               <image>nexus.terrestris.de/shogun/${project.artifactId}:latest</image>
               <tags>
                 <tag>${project.version}</tag>
+                <!--suppress UnresolvedMavenProperty -->
+                <tag>${project.version}-${git.commit.id.abbrev}</tag>
               </tags>
             </to>
             <container>
--- a/pom.xml
+++ b/pom.xml
@@ -223,14 +223,12 @@
               <image>nexus.terrestris.de/shogun/${project.artifactId}:latest</image>
               <tags>
                 <tag>${project.version}</tag>
-                <!--suppress UnresolvedMavenProperty -->
-                <tag>${project.version}-${git.commit.id.abbrev}</tag>
               </tags>
             </to>
             <container>
               <jvmFlags>-Djava.security.egd=file:/dev/./urandom</jvmFlags>
               <jvmFlags>-Dspring.config.location=/config/application.yml</jvmFlags>
-              <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
+              <creationTime>EPOCH</creationTime>
             </container>
           </configuration>
         </plugin>
--- a/shogun-config/pom.xml
+++ b/shogun-config/pom.xml
@@ -72,14 +72,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>com.google.cloud.tools</groupId>
-        <artifactId>jib-maven-plugin</artifactId>
-        <configuration>
-          <!-- we don't want jib to execute on this module since its a library -->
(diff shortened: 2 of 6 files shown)


shogun-config/pom.xml
```diff
@@ -72,6 +72,14 @@
...
         </executions>
       </plugin>
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>jib-maven-plugin</artifactId>
+        <configuration>
+          <!-- we don't want jib to execute on this module since its a library -->
+          <skip>true</skip>
+        </configuration>
+      </plugin>
```
dnlkoch, line 82: The plugin will run only if it's explicitly configured for the module. So this block is not needed.

shogun-lib/pom.xml
```diff
@@ -19,11 +19,18 @@
...
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>jib-maven-plugin</artifactId>
+        <configuration>
+          <!-- we don't want jib to execute on this module since its a library -->
+          <skip>true</skip>
+        </configuration>
+      </plugin>
```
dnlkoch, line 29: See above.

shogun-manager/pom.xml
```diff
@@ -35,4 +35,17 @@
...
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>jib-maven-plugin</artifactId>
+        <configuration>
+          <!-- we don't want jib to execute on this module since its a library -->
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
```
dnlkoch, line 49: See above.

pom.xml
```diff
@@ -217,20 +217,18 @@
...
             <from>
-              <image>openjdk:11-slim</image>
+              <image>openjdk@sha256:2b6eb2c178908b6704d82258652806eca3c0cd9261df7f5edcc3486d63bc6e60</image>
             </from>
             <to>
               <image>nexus.terrestris.de/shogun/${project.artifactId}:latest</image>
               <tags>
                 <tag>${project.version}</tag>
-                <!--suppress UnresolvedMavenProperty -->
-                <tag>${project.version}-${git.commit.id.abbrev}</tag>
```
dnlkoch, line 227: Any specific reason why you removed this? It was introduced intentionally.

ahennr: Since I've found no proper configuration and a `UnresolvedMavenProperty` was suppressed, I Thought it could be removed.

Or is the config of `spring-boot` uses in SHOGun here?

shogun-config/pom.xml
ahennr: JIB will run on each submodule defined in SHOGun's parent pom IMHO.
You can see this in output of `mvn clean compile jib:build` and in the plugins multi-module [example](https://github.com/GoogleContainerTools/jib/tree/master/examples/multi-module)

As far as I understand, (shared) libraries that are do not run as a standalone application need to be skipped from `jib:build` step. Maybe we need to introduce the plugin for reproducible builds as well,m what do you think?

dnlkoch: According to the logs from the last (stable) [build](https://github.com/terrestris/shogun/runs/1279433772?check_suite_focus=true) the execution will be applied to the desired modules only and I assume it's because the `jib` build process is bound to the `package`/`deploy` phases. So `mvn clean compile jib:build` is not necessarily needed / intended to be run.

pom.xml
dnlkoch: Yes, this element is a default for every submodule (where the plugin is being used, e.g. `shogun-boot`) and the (unresolved) `git` properties will be made available for each one via the git plugin - this is why it's suppressed here.

ahennr: ok

shogun-config/pom.xml
ahennr: I've removed the unneeded config in 915aa882

shogun-lib/pom.xml
ahennr: I've removed the unneeded config in 915aa882

shogun-manager/pom.xml
ahennr: I've removed the unneeded config in 915aa882

ahennr:
Can this be merged?

dnlkoch:
Yes, please merge.

PR #152, opened 2020-10-29
Fix current build and warnings

dnlkoch:
This provides some minor patches to:
  * fix the current build (see 71e1f58, also [here](https://docs.spring.io/spring-boot/docs/2.0.1.RELEASE/maven-plugin/examples/repackage-classifier.html) and [here](https://stackoverflow.com/questions/59900637/error-testengine-with-id-junit-vintage-failed-to-discover-tests-with-spring) for more details).
  * fix a warning about usage of a deprecated class/static (see 5b1ec7e).
  * fix the test suite by excluding `junit-vintage-engine`, which is not compatible with the latest versions of `junit` (see a781a36).


KaiVolland (approved):
Nice. Please merge.

ahennr (commented):
:top:

PR #165, opened 2020-11-13
implement permission check by target entity id

LukasLohoff:
This implements a new method 

```java
boolean hasPermission(User user, Long entityId, String targetDomainType, PermissionType permission);
```
in the EntityPermissionEvaluator which can be used to evaluate permissions based on the target entity id rather than providing the full entity.

Example usage:
```java
@PreAuthorize("hasPermission(filterObjectId, 'de.terrestris.shogun.model.User', 'READ')")
```


dnlkoch (approved):
Awesome @LukasLohoff!

PR #172, opened 2020-11-23
Introduces updatePartial interface

### LINKED ISSUE #169: Support for partial updates
We should add interfaces for partial updates of entities.

[HTTP `PATCH`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) should be appropriate for this case.

Currently if we have large objects or objects with read-only fields we need to send huge objects and do some crazy workarounds.
When changing just one attribute of an entity, there should be an interface to post only this one changed attribute.

KaiVolland:
This introduces controller and service methods to partially update entities.

The passed values object has to include the `id` which has to be equal to the one from the PATCH url.

Values which can't be associated with the entity will be ignored:

#### Valid :heavy_check_mark: 
Send [HTTP PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) request to: `http://localhost:9090/shogun-boot/applications/21`
**_200 OK_**
```
{
  "id": 21,
  "name": "My new App"
}
```

#### Valid :heavy_check_mark: 
Send [HTTP PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) request to: `http://localhost:9090/shogun-boot/applications/21`
**_200 OK_**
```
{
  "id": 21,
  "name": "My new App",
  "opacity": 0.234
}
```

#### Invalid :x:  
Send [HTTP PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) request to: `http://localhost:9090/shogun-boot/applications/21`
**_400 Bad Request_**
```
{
  "id": 22,
  "name": "My new App"
}
```

#### Invalid :x:  
Send [HTTP PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) request to: `http://localhost:9090/shogun-boot/applications/21`
**_400 Bad Request_**
```
{
  "name": "My new App"
}
```

Closes #169

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
@@ -16,6 +16,7 @@ import org.springframework.web.bind.annotation.*;
 import org.springframework.web.server.ResponseStatusException;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 // TODO Specify and type extension of BaseService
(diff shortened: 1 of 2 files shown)


dnlkoch (approved):

LukasLohoff:
You could also catch `NumberFormatException` when values.get("id") is not parseable. Other than that, this looks very solid!

LukasLohoff (approved):

marcjansen:
Really cool.

Some questions, which someone might answer:

* Will the `modified` field get set with the correct date when a partial update happens?
* Is it possible to nullify fields when previously they had a value? E.g.:

PATCH to http://localhost:9090/shogun-boot/applications/21
```
{
  "id": 21,
  "som_nullable_field": null
}
```

* You write "values which can't be associated with the entity will be ignored", will this somehow be reported? Or logged? I could also think of failing the transaction loudly, when unexpected fields arrive, would make sense to me.

marcjansen:
Another thing which comes to my mind:

Is it possible to model `Map<String, Object> values` in a way that Spring already does the casting to the right data-types? We know that it should always look like this (fantasy code):

```
Map
  key "id" -> value of type int/long
  n further keys -> value of type Object 
```

I have no clue whether Java can express this somehow, maybe with (extending) a POJO somehow? I really do not know how one could do this.

KaiVolland:
> * Will the `modified` field get set with the correct date when a partial update happens?

The code is almost the same as the code of the PUT update method. The modified field is not handled specially as it is handled via Hibernate annotation `@UpdateTimestamp` at the `BaseEntity`. 
> * Is it possible to nullify fields when previously they had a value? E.g.:
> 
> PATCH to http://localhost:9090/shogun-boot/applications/21
> 
> ```
> {
>   "id": 21,
>   "som_nullable_field": null
> }
> ```

Yes. Just like in the PUT update method.

> * You write "values which can't be associated with the entity will be ignored", will this somehow be reported? Or logged? I could also think of failing the transaction loudly, when unexpected fields arrive, would make sense to me.

This is neither reported nor logged. I have no strong opinion on this but simply ignoring seems to work and keeps the code small and clean.

> * Is it possible to model Map<String, Object> values in a way that Spring already does the casting to the right data-types? We know that it should always look like this (fantasy code):

JAVA itself pretty sure can't do this. Maybe there are some Kotlin (or similar) things that can do this. But i don't think its even necessary as we transfer it to a JSON-Object right before persisting anyway.

ahennr (approved):

buehner (approved):
LGTM

hwbllmnn (approved):

PR #178, opened 2020-12-01
Filter out null values before mapping them to a list

dnlkoch:
This fixes any potential NPE.

Not relevant for the `master` since the service is not needed in newer versions.


MrSnyder:
LGTM, please merge

PR #180, opened 2020-12-01
Introduce conditional controller instances (v3)

dnlkoch:
This makes the creation of all controller instances configurable, e.g. via the following entries in the `application.properties`:

```
controller.applications.enabled=false
controller.files.enabled=true
controller.groups.enabled=false
controller.identities.enabled=false
controller.imagefiles.enabled=true
controller.layers.enabled=true
controller.roles.enabled=false
controller.users.enabled=true
```

Once accepted, this should also be applied to the `master` branch.


marcjansen (commented):
Very nice

hwbllmnn (approved):

PR #189, opened 2020-12-07
Add git commit hash to build information

annarieger:
Add `commithash` property to `build-info` config to grab the commit hash info while packaging and make it accessible via `build.properties` file.

```
build.artifact=shogun-boot
build.commithash=2f454564
build.encoding.reporting=UTF-8
build.encoding.source=UTF-8
build.group=de.terrestris
build.name=SHOGun Boot
build.time=2020-12-07T12\:17\:33.402Z
build.version=6.1.1-SNAPSHOT
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,7 @@
                 <additionalProperties>
                   <encoding.source>${project.encoding}</encoding.source>
                   <encoding.reporting>${project.encoding}</encoding.reporting>
+                  <commithash>${git.commit.id.abbrev}</commithash>
                 </additionalProperties>
               </configuration>
             </execution>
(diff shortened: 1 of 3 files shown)


hwbllmnn (approved):

dnlkoch:
Note: This information is available via the actuator interface `actuator/info`.

dnlkoch (approved):

annarieger:
> Note: This information is available via the actuator interface `actuator/info`.

Since actuator interface is available for `ADMIN` role [only](https://github.com/terrestris/shogun/blob/master/shogun-boot/src/main/java/de/terrestris/shogun/boot/config/BootWebSecurityConfig.java#L39), providing these infos via `app/info` will enable access for other user groups as well.

dnlkoch:
> Since actuator interface is available for ADMIN role only, providing these infos via app/info will enable access for other user groups as well.

Yes, you're right, but it can be adjusted to your needs in the project specific web security config.

PR #190, opened 2020-12-08
Adds GraphQL support for File and ImageFile

KaiVolland:
This adds GraphQL support for `File` and `ImageFile` classes.

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/graphql/GraphQLDataFetchers.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/graphql/GraphQLDataFetchers.java
@@ -1,8 +1,6 @@
 package de.terrestris.shogun.lib.graphql;
 
-import de.terrestris.shogun.lib.service.ApplicationService;
-import de.terrestris.shogun.lib.service.LayerService;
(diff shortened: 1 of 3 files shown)


shogun-lib/src/main/java/de/terrestris/shogun/lib/graphql/GraphQLDataFetchers.java
```diff
@@ -19,36 +17,65 @@
...
             return this.userService.findOne(userId);
         };
     }
-
     public DataFetcher getAllUsers() {
         return dataFetchingEnvironment -> this.userService.findAll();
     }
+
+    // File
+    public DataFetcher getFileById() {
```
dnlkoch, line 60: The REST controllers for file entities are accepting the UUID of the file instead of the ID. Not blocking, but I think we should harmonize this behaviour.

dnlkoch (approved):

shogun-lib/src/main/java/de/terrestris/shogun/lib/graphql/GraphQLDataFetchers.java
KaiVolland: 99b46e9

PR #199, opened 2020-12-15
Group permission retrieval (v3)

dnlkoch:
This suggests to apply some minor changes, especially for easier handling of group permissions in child projects by:

* Refactoring the `BasePermissionEvaluator` and `BaseEntityPermissionEvaluator` to become easier reusable and extendable.
* Adding the method `findFor(BaseEntity entity, Group group, User user)` to both the `GroupInstancePermission`- and `GroupClassPermissionService` for taking the group membership of a user into account while getting the appropriate permissions of a user.

Assorted changes might be applied to the `main` as soon as accepted here.

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shoguncore/service/security/permission/GroupInstancePermissionService.java
+++ b/shogun-lib/src/main/java/de/terrestris/shoguncore/service/security/permission/GroupInstancePermissionService.java
@@ -12,14 +12,11 @@ import de.terrestris.shoguncore.service.BaseService;
 import de.terrestris.shoguncore.service.security.IdentityService;
 import de.terrestris.shoguncore.specification.security.permission.GroupInstancePermissionSpecifications;
 import de.terrestris.shoguncore.specification.security.permission.PermissionCollectionSpecification;
-import javassist.NotFoundException;
+import java.util.List;
+import java.util.Optional;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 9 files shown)


shogun-lib/src/main/java/de/terrestris/shoguncore/service/security/permission/GroupInstancePermissionService.java
```diff
@@ -53,24 +50,41 @@
         ));
     }
 
-    public PermissionCollection findPermissionCollectionFor(BaseEntity entity, Group group) {
-        Optional<GroupInstancePermission> groupInstancePermission = this.findFor(entity, group);
+    public Optional<GroupInstancePermission> findFor(BaseEntity entity, Group group, User user) {
```
KaiVolland, line 53: Could you add some small code documentation here? Maybe also to the other methods.

KaiVolland (approved):
Nice stuff. Some more documentation would be cool though.

shogun-lib/src/main/java/de/terrestris/shoguncore/service/security/permission/GroupInstancePermissionService.java
dnlkoch: You're right, documentation is lacking. I'll add it.

dnlkoch: Added in e134a14.

PR #203, opened 2020-12-17
implement setPermission for a list of entities

### LINKED ISSUE #184: speed up setPermission methods
I'm trying to create ~20000 new entities and also set permissions for every one. Unfortunately, `GroupInstancePermissionService::setPermission` takes about 3 seconds per permission.

See https://github.com/terrestris/shogun/blob/7d5df50f6026bded33a071a5e77333e956ec5a00/shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/UserInstancePermissionService.java#L67 for example

Ideas:

- is the check for existing permissions necessary (especially for new entities)? could this be optional maybe?
- we could provide a method `setPermission(List<BaseEntity> persistedEntities, Group group, PermissionCollectionType pct)` where `repository.saveAll()` can be used to create many permissions for a single group

Other suggestions are welcome…

LukasLohoff:
Resolves #184:

- introduces
   -  `setPermission(List<BaseEntity> entityList, Group group, PermissionCollectionType pct)` and
   - `setPermission(List<BaseEntity> entityList, User user, PermissionCollectionType pct)` 
- this can be used to set a permission for a collection of entities at once
- makes use of `repository.saveAll()` and should speed up permission creation considerably
- existing duplicate permissions are still checked and removed if present
- didn't do any detailed performance testing, but the import mentioned in #184 now only takes a few minutes (including permissions)

Example usage:

```java
@Autowired
protected GroupInstancePermissionService groupInstancePermissionService;

@Autowired
protected LayerRepository layerRepository;

Group group = ...
List<Layer> layers = layerRepository.findAll();

groupInstancePermissionService.setPermission(layers, group, PermissionCollectionType.ADMIN);
```

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/GroupInstancePermissionService.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/GroupInstancePermissionService.java
@@ -128,7 +128,7 @@ public class GroupInstancePermissionService extends BaseService<GroupInstancePer
      * @param group The SHOGun group
      * @param permissionCollectionType The permission collection type (e.g. READ, READ_WRITE)
      */
-    public void setPermissionBulk(
+    public void setPermission(
         List<? extends BaseEntity> persistedEntityList,
         Group group,
         PermissionCollectionType permissionCollectionType
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/GroupInstancePermissionService.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/GroupInstancePermissionService.java
@@ -112,7 +112,7 @@ public class GroupInstancePermissionService extends BaseService<GroupInstancePer
             throw new RuntimeException("Could not find requested permission collection");
         }
 
-        checkForExistingPermissions(group, permissionCollection.get(), persistedEntity);
+        clearExistingPermission(group, permissionCollection.get(), persistedEntity);
 
         GroupInstancePermission groupInstancePermission = new GroupInstancePermission();
         groupInstancePermission.setGroup(group);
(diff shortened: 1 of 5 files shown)


shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/GroupInstancePermissionService.java
```diff
@@ -111,26 +112,61 @@ public void setPermission(BaseEntity persistedEntity, Group group, PermissionCol
...
+        repository.save(groupInstancePermission);
+    }
+
+    /**
+     * Set Permission for SHOGun group for multiple entities at once
+     * @param persistedEntityList A collection of entities to set permission for
+     * @param group The SHOGun group
+     * @param permissionCollectionType The permission collection type (e.g. READ, READ_WRITE)
+     */
+    public void setPermissionBulk(
```
dnlkoch, line 131: Non blocking, but in terms of naming harmonization I would suggest to rename the method to `setPermission`.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/UserInstancePermissionService.java
```diff
@@ -102,6 +92,52 @@ public void setPermission(BaseEntity persistedEntity, User user, PermissionColle
         repository.save(userInstancePermission);
     }
 
+    /**
+     * Set Permission for SHOGun user for multiple entities at once
+     * @param persistedEntityList A collection of entities to set permission for
+     * @param user The user
+     * @param permissionCollectionType The permission collection type (e.g. READ, READ_WRITE)
+     */
+    public void setPermissionBulk(
```
dnlkoch, line 101: See note about naming above.

dnlkoch (approved):
Very nice @LukasLohoff!

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/GroupInstancePermissionService.java
LukasLohoff: fixed

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/permission/UserInstancePermissionService.java
LukasLohoff: also fixed

LukasLohoff:
Thanks! Renamed the methods as suggested

PR #207, opened 2020-12-17
Create github action to test if application starts up

simonseyock:
This adds a github action that simply tests if the application starts up to catch the most severe errors.

Here is how a successful run looks like: https://github.com/simonseyock/shogun/runs/1571654866?check_suite_focus=true
and here is a not successful one: https://github.com/simonseyock/shogun/runs/1571697016?check_suite_focus=true

Note that the docker-compose logs and maven output are only displayed at the end and if the run was not successful. The latter could be changed so it is shown always, but still at the end.

### DIFF
--- a/.github/workflows/startup.yml
+++ b/.github/workflows/startup.yml
@@ -1,0 +1,57 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Startup Shogun
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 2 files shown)


simonseyock:
#208 should get merged first

simonseyock:
@terrestris/devs can this get merged?

KaiVolland (approved):

dnlkoch (approved):
Very nice!

Just wondering, is it somehow possible to access the logs of a successful startup?

simonseyock:
the logs are now shown always

marcjansen:
Thanks

PR #209, opened 2020-12-17
Use an Annotation to mark graph ql queries.

simonseyock:
This introduces an annotation `@GraphQLQuery` that annotates a method that returns a `DataFetcher` for a graph ql query. The GraphQLProvider will search for all methods with this annotation automatically on application start up.

Usage:
```
    @GraphQLQuery(name = "allUsers")
    public DataFetcher getAllUsers() {
        return dataFetchingEnvironment -> this.userService.findAll();
    }
```

It can be used without the name parameter in which case the method name is used.

It has to be checked that the GraphQLProvider will correctly search for the annotation if used from a child project.

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/graphql/GraphQLDataFetchers.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/graphql/GraphQLDataFetchers.java
@@ -26,68 +26,80 @@ public class GraphQLDataFetchers {
     ImageFileService imageFileService;
 
     // Application
+    @GraphQLQuery(name = "applicationById")
     public DataFetcher getApplicationById() {
         return dataFetchingEnvironment -> {
             Long applicationId = ((Integer) dataFetchingEnvironment.getArgument("id")).longValue();
             return this.applicationService.findOne(applicationId);
         };
     }
+    @GraphQLQuery(name = "allApplications")
     public DataFetcher getAllApplications() {
         return dataFetchingEnvironment -> this.applicationService.findAll();
     }
 
     // Layer
+    @GraphQLQuery(name = "layerById")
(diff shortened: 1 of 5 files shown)


shogun-lib/src/main/resources/graphql/shogun.graphqls
```diff
@@ -11,6 +11,7 @@ type Query {
     fileById(id: Int): File
     fileByUuid(uuid: String): File
     allFiles: [File]
+    getImageFileByUuid(uuid: String): File
```
KaiVolland, line 14: Please remove the `get` to be consistent.

KaiVolland:
Very nice!

shogun-lib/src/main/java/de/terrestris/shogun/lib/graphql/GraphQLDataFetchers.java
```diff
@@ -85,7 +85,7 @@ public DataFetcher getAllFiles() {
     }
 
     // ImageFile
-    @GraphQLQuery(name = "getImageFileByUuid")
+    @GraphQLQuery(name = "imageFileByUuid")
```
KaiVolland, line 88: You missed to add the change to the schema file. :)

simonseyock: true

shogun-lib/src/main/resources/graphql/shogun.graphqls
simonseyock, line 14:
```suggestion
    imageFileByUuid(uuid: String): File
```

KaiVolland (approved):
Thanks. Please merge.

PR #213, opened 2020-12-22
Ensure the created timestamp is not going to be overridden (v3)

dnlkoch:
This ensures the `created` timestamp is not going to be overridden in REST PUT (update) if it's not given in the PUT body.

In addition some logger hints are added in context of the permission retrieval for users that are member of multiple groups.

This is also candidate for the `main` once accepted.


hwbllmnn (approved):

annarieger (approved):

LukasLohoff (approved):

dnlkoch:
Thanks for the reviews!

PR #221, opened 2021-01-08
Introduce a generic GraphQL base resolver (v3)

dnlkoch:
This introduces a `BaseGraphQLDataFetcher` which registers several base GraphQL queries and mutations (!) once implemented for a given entity (e.g. `ApplicationGraphQLDataFetcher`). The names of the queries will be autogenerated for each available operation following simple conventions:

* `findAll`
  * `all{BaseEntities}`, e.g. `allApplications`
* `findOne`
  * `{BaseEntity}ById`, e.g. `applicationById`
* `create`
  * `create{BaseEntity}`, e.g. `createApplication`
* `update`
  * `update{BaseEntity}`, e.g. `updateApplication`
* `delete`
  * `delete{BaseEntity}`, e.g. `deleteApplication`

Please see the `shogun.graphqls` for a full list of available queries and mutations.

Upgrade notes for projects:

* The property `shogun.graphql.skipBean` is no longer needed and can be removed from the `application.properties`
* Any custom `GraphQLProvider` can be removed.
* Any custom data fetcher (registered in the `GraphQLProvider`) should be moved to a single resolver, e.g.:
```java
@Component
public class MyEntityGraphQLDataFetcher extends BaseGraphQLDataFetcher<MyEntity, MyEntityService> {

    @Autowired
    private MyEntityRepository repository;

    @GraphQLQuery(name = "myEntityBySomething")
    public DataFetcher myEntityBySomething() {
        return dataFetchingEnvironment -> {
            String something = dataFetchingEnvironment.getArgument("something");
            return this.repository.findBySomething(something);
        };
    }
}
```
(Note the annotation `@GraphQLQuery` which defines the name of the query which is needed to register it in the schema file.)

Again, this should be applied to the `main` as well once accepted.

(In addition a new service method for returning the owner an entity has been added.)


shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/GraphQLProvider.java
```diff
@@ -30,29 +40,19 @@
...
 
-    @Autowired
-    GraphQLDataFetchers graphQLDataFetchers;
-
     @Bean
-    @ConditionalOnProperty(
-        value="shogun.graphql.skipBean",
-        havingValue = "false",
-        matchIfMissing = true
-    )
```
KaiVolland, line 41: Did you test this with child projects?

KaiVolland:
This looks great!

KaiVolland (approved):

dnlkoch:
Sorry, I should have stated this more clearly. Yes, a single resolver can contain multiple fetchers and it should be grouped by the given base entity (similar to the services and friends) - no need to create a single class for each fetcher.

dnlkoch:
Thanks for the reviews!

PR #223, opened 2021-01-14
Add interfaces for jsonb fields

simonseyock:
**This is a breaking change**

This introduces interfaces for the jsonb types. They only extend `Serializable` and are very flexible this way. `ApplicationClientConfig` is changed to be an interface as well.

This allows a strict typing of the jsonb fields of `User`, `Application` and `Layer` by implementing the interfaces. The typing can be done inside of a (child) project using shogun.

The available interfaces are:
* `ApplicationClientConfig` for field `Application#clientConfig`
* `ApplicationLayerConfig` for field `Application#layerConfig`
* `ApplicationToolConfig` for field `Application#toolConfig`
* `ApplicationLayerTree` for field `Application#layerTree`
* `LayerClientConfig` for field `Layer#clientConfig`
* `LayerFeature` for the elements of field `Layer#features` (it is a `List<LayerFeature>`)
* `LayerSourceConfig` for the field `Layer#sourceConfig`
* `UserClientConfig` for the field `User#clientConfig`
* `UserDetails` for the field `User#details`

For the deserialization it is necessary to annotate the implementing class with `JsonSuperType` and `JsonDeserialize`,  see examples below. Also it is recommended to add `@ToString` and `@EqualsAndHashCode` for any type that is stored in a jsonb field to the database. As the reflections library outputs very many warnings, it is advised to add the following lines to `log4j2.properties`:
```
logger.reflections.name=org.reflections
logger.reflections.level=error
```

It can be used for example with the `UserDetails` like this:

`myproject/model/jsonb/UserDetails.java`
```java
@Data
@ToString
@EqualsAndHashCode
    private boolean  someFlag;
    private String text;
}
```

And then it can be used like this:
```java
    UserDetails details = (UserDetails) user.getDetails();
    if (details == null) {
        details = new UserDetails();
        user.setDetails(details);
    }
    details.setText("some text");
    userService.update(user.getId(), user);
```

**Attention**
1. This does not prevent the developer to implement 2 different types that implement `UserDetails`. This should be avoided.
2. If a parent project already implements a `UserDetails` this one should be extended. *TODO: it is not clear how this will affect `JsonSuperType`*
3. This drops the `logoPath` and `mapView` from `ApplicationClientConfig`. If needed they need to be added inside the child projects specifically.


shogun-lib/pom.xml
```diff
@@ -221,6 +221,11 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.reflections</groupId>
+      <artifactId>reflections</artifactId>
+      <version>0.9.12</version>
```
dnlkoch, line 227: Please define dependency in parent pom.

shogun-lib/src/main/java/de/terrestris/shoguncore/annotation/JsonSuperType.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.io.Serializable;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface JsonSuperType {
```
dnlkoch, line 11: A short documentation about the usage of this annotation could be very helpful.

shogun-lib/src/main/java/de/terrestris/shoguncore/config/JacksonConfig.java
```diff
@@ -1,19 +1,67 @@
...
+    public ObjectMapper get() {
+        return objectMapper();
+    }
+
+    @PostConstruct
+    public static void init() {
+        if (mapper == null) {
+            mapper = new ObjectMapper().registerModule(jtsModule());
+
+            mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
dnlkoch, line 49: According to the [docs](https://docs.spring.io/spring-boot/docs/2.4.2/reference/html/howto.html#howto-customize-the-jackson-objectmapper) the default mapper created by spring has also set/disabled the following features:

* `MapperFeature.DEFAULT_VIEW_INCLUSION`
* `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES`
* `SerializationFeature.WRITE_DATES_AS_TIMESTAMPS`

If they aren't disabled by default anyway, I would suggest to also disable them.

shogun-lib/src/main/java/de/terrestris/shoguncore/model/jsonb/ApplicationClientConfig.java
```diff
@@ -5,9 +5,6 @@
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-@Data
-@AllArgsConstructor
-@NoArgsConstructor
-public class ApplicationClientConfig implements Serializable {
-    private String logoPath;
+public interface ApplicationClientConfig extends Serializable {
+
```
dnlkoch, line 9:
```suggestion
```

shogun-lib/src/main/java/de/terrestris/shoguncore/model/jsonb/UserDetails.java
```diff
@@ -0,0 +1,7 @@
+package de.terrestris.shoguncore.model.jsonb;
+
+import java.io.Serializable;
+
+public interface UserDetails extends Serializable  {
+
```
dnlkoch, line 6:
```suggestion
```

shogun-lib/src/main/java/de/terrestris/shoguncore/model/jsonb/ApplicationClientConfig.java
```diff
@@ -5,9 +5,6 @@
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-@Data
-@AllArgsConstructor
-@NoArgsConstructor
-public class ApplicationClientConfig implements Serializable {
-    private String logoPath;
+public interface ApplicationClientConfig extends Serializable {
```
dnlkoch, line 8: Can we also introduce the types for all other JSONB fields on the `Application`?

dnlkoch (commented):
Absolutely awesome so far! I just noted some minor remarks you may want to address.

simonseyock:
Thanks for the review! I adressed your remarks

shogun-lib/src/main/java/de/terrestris/shoguncore/model/Application.java
```diff
@@ -41,17 +44,15 @@
...
     @Basic(fetch = FetchType.LAZY)
-    private Map<String, Object> layerConfig;
+    private LayerConfig layerConfig;
 
     @Type(type = "jsonb")
     @Column(columnDefinition = "jsonb")
     @Basic(fetch = FetchType.LAZY)
-    private Map<String, Object> toolConfig;
-
+    private LayerToolConfig toolConfig;
```
dnlkoch, line 57:
```suggestion
    private ApplicationToolConfig toolConfig;
```

pom.xml
```diff
@@ -583,6 +583,12 @@
         <artifactId>jts-core</artifactId>
         <version>${jts.version}</version>
       </dependency>
+
+      <dependency>
+        <groupId>org.reflections</groupId>
+        <artifactId>reflections</artifactId>
+        <version>0.9.12</version>
```
dnlkoch, line 590: Could you also introduce a variable for the version?

shogun-lib/src/main/java/de/terrestris/shoguncore/model/Application.java
simonseyock: I prefixed all types that reside inside Application with `Application`

dnlkoch (approved):

PR #225, opened 2021-01-14
Add interfaces for jsonb fields

simonseyock:
**This is a breaking change**

This introduces interfaces for the jsonb types. They only extend `Serializable` and are very flexible this way. `ApplicationClientConfig` is changed to be an interface as well.

This allows a strict typing of the jsonb fields of `User`, `Application` and `Layer` by implementing the interfaces. The typing can be done inside of a depending project of shogun.

The available interfaces are:
* `ApplicationClientConfig` for field `Application#clientConfig`
* `LayerClientConfig` for field `Layer#clientConfig`
* `LayerFeature` for the elements of field `Layer#features` (it is a `List<LayerFeature>`)
* `LayerSourceConfig` for the field `Layer#sourceConfig`
* `UserClientConfig` for the field `User#clientConfig`
* `UserDetails` for the field `User#details`

For the deserialization it is necessary to annotate the implementing class with `JsonSuperType` and `JsonDeserialize`,  see examples below. Also it is recommended to add `@ToString` and `@EqualsAndHashCode` for any type that is stored in a jsonb field to the database. As the reflections library outputs very many warnings, it is advised to add the following lines to `log4j2.properties`:
```
logger.reflections.name=org.reflections
logger.reflections.level=error
```

It can be used for example with the `UserDetails` like this:

`myproject/model/jsonb/UserDetails.java`
```java
@Data
@ToString
@EqualsAndHashCode
    private boolean  someFlag;
    private String text;
}
```

And then it can be used like this:
```java
    UserDetails details = (UserDetails) user.getDetails();
    if (details == null) {
        details = new UserDetails();
        user.setDetails(details);
    }
    details.setText("some text");
    userService.update(user.getId(), user);
```

**Attention**
1. This does not prevent the developer to implement 2 different types that implement `UserDetails`. This should be avoided.
2. If a parent project already implements a `UserDetails` this one should be extended.
3. This drops the `logoPath` and `mapView` from `ApplicationClientConfig`. If needed they need to be added inside the child projects specifically.


simonseyock:
TODO: find out if injected spring values actually work statically. Is there a working way to do this?

Problem is that hibernate seems to request the object mapper before spring is initiated.

shogun-lib/pom.xml
```diff
@@ -231,11 +231,6 @@
       <artifactId>HikariCP</artifactId>
       <scope>test</scope>
     </dependency>
-      <dependency>
-          <groupId>org.reflections</groupId>
-          <artifactId>reflections</artifactId>
-          <version>0.9.12</version>
-      </dependency>
```
dnlkoch, line 238: This should be readded (without the version).

simonseyock: this is now in the pom.xml on top level

simonseyock: nevermind, i add it

dnlkoch (approved):

KaiVolland:
> ~TODO: find out if injected spring values actually work statically. Is there a working way to do this?~ Done
> 
> Problem is that hibernate seems to request the object mapper before spring is initiated.

This is still an issue and leads to two different instances (configurations) of the Objectmapper. The one used by hibernate is missing the SRID value for the JTS module.

PR to fix this is in work here: https://github.com/terrestris/shogun/pull/266

PR #232, opened 2021-01-19
Add missing type READ_UPDATE_DELETE (v3)

dnlkoch:
This adds the missing `READ_UPDATE_DELETE` to the `PermissionCollectionType` enum.

**Note:** To make use of this type, existing projects must make it available via an own db migration as adding them to the migrations of shogun-boot might not work as it could be out of order with the ones defined in the project:

```sql
INSERT INTO permissions VALUES (
    nextval('hibernate_sequence'),
    NOW()::timestamp,
    NOW()::timestamp,
    'READ_UPDATE_DELETE'
);
INSERT INTO permission VALUES (
    (SELECT id FROM permissions WHERE id = currval('hibernate_sequence')),
    'READ'
);
INSERT INTO permission VALUES (
    (SELECT id FROM permissions WHERE id = currval('hibernate_sequence')),
    'UPDATE'
);
INSERT INTO permission VALUES (
    (SELECT id FROM permissions WHERE id = currval('hibernate_sequence')),
    'DELETE'
);
```

*TODO:* We should define a better convention for handling such scenarios!


annarieger (approved):

dnlkoch:
Thanks for the review!

PR #233, opened 2021-01-19
Several permission service and repository features (v3)

dnlkoch:
This applies some features to the permission services and repositories:

* Add method `deleteAllByUserId()` to delete all use instance and class permissions via the ID of the user. This performances much faster than getting the appropriate permissions first in scenarios where a lot of permissions are set.
* Add method `findFor()` for finding the group instance permissions by entity.
* Log the IDs of the handled entities only to minimize log footprint.


simonseyock (approved):

Unraveler:
Looks good!

dnlkoch:
Thanks for the reviews!

PR #239, opened 2021-01-25
Allow JsonSuperType annotation on multiple types that extend each other (v3)

simonseyock:
This PR is on the v3 branch.

This commit solves the problem of having multiple implementations of one of the JSONB type.

Imagine we have this project structure: `shogun-lib > project a > project b` and we extend the type `UserDetails` on project a and on project b. Then we would want to extend the UserDetails from project a in project b. But both classes would have the `JsonSuperType` annotation.

This PR will now check if multiple types deserialize to the same super type and will choose the type that is the furthest down the implementation chain. In the example that would be the `UserDetails` from project b.


dnlkoch (approved):
Looks pretty useful! Thanks for this!

PR #242, opened 2021-01-26
Added DateTime GraphQL Scalar definition

Unraveler:
- Added new DateTime scalar definition for GraphQL
- Added setDateTimeConfig to Jackson Config objectMapper 

please review @terrestris/devs

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shoguncore/config/JacksonConfig.java
+++ b/shogun-lib/src/main/java/de/terrestris/shoguncore/config/JacksonConfig.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.util.ISO8601DateFormat;
 import com.vladmihalcea.hibernate.type.util.ObjectMapperSupplier;
 import de.terrestris.shoguncore.annotation.JsonSuperType;
 import org.locationtech.jts.geom.GeometryFactory;
@@ -18,6 +19,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 import javax.annotation.PostConstruct;
+import java.text.SimpleDateFormat;
 import java.util.Set;
 
 @Configuration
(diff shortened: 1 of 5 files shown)


shogun-lib/src/main/java/de/terrestris/shoguncore/config/JacksonConfig.java
```diff
@@ -5,6 +5,7 @@
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.util.ISO8601DateFormat;
```
dnlkoch, line 8: I think this line can be removed.

shogun-lib/src/main/java/de/terrestris/shoguncore/model/BaseEntity.java
```diff
@@ -1,5 +1,6 @@
 package de.terrestris.shoguncore.model;
 
+import com.fasterxml.jackson.annotation.JsonFormat;
```
dnlkoch, line 3: Can also be removed.

dnlkoch (approved):

shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/DateTimeScalar.java
```diff
@@ -0,0 +1,47 @@
...
+            return parseDateFromAstLiteral(input);
+        }
+    });
+
+    private static Object serializeDate(Object dataFetcherResult) {
+        return objectMapper().convertValue(dataFetcherResult, String.class);
+    }
+
+    private static Object parseDateFromVariable(Object dataFetcherResult) {
+        return objectMapper().convertValue(dataFetcherResult, String.class);
```
simonseyock, line 41: i think this should be like this:
```
if (input instanceof String) {
                return objectMapper.readValue((String) input, Date.class);
            }
            throw new CoercingParseValueException("...");
```

shogun-lib/src/main/java/de/terrestris/shoguncore/graphql/scalar/DateTimeScalar.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    private static Object parseDateFromAstLiteral(Object dataFetcherResult) {
+        return objectMapper().convertValue(dataFetcherResult, String.class);
```
simonseyock, line 45: this should be like this:
```
if (input instanceof StringValue) {
    return objectMapper.readValue(((StringValue) input).getValue(), Date.class);
}
throw new CoercingParseValueException("...");
```

PR #244, opened 2021-01-26
Fix call of `hasPermission` of updatePartial for users not having ROLE_ADMIN

ahennr:
Title says it all.

Plz review @terrestris/devs

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
@@ -472,10 +472,9 @@ public abstract class BaseController<T extends BaseService<?, S>, S extends Base
                 throw new ResponseStatusException(HttpStatus.BAD_REQUEST);
             }
 
-            Optional<S> persistedEntity = service.findOne(entityId);
-
-            if (persistedEntity.isPresent()) {
-                S updatedEntity = service.updatePartial(entityId, values);
+            S persistedEntity = service.findOne(entityId).orElseThrow();
+            if (persistedEntity != null) {
+                S updatedEntity = service.updatePartial(entityId, persistedEntity, values);
 
                 LOG.trace("Successfully updated values for entity of type {} with ID {}",
                     getGenericClassName(), entityId);
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
@@ -463,7 +463,7 @@ public abstract class BaseController<T extends BaseService<?, S>, S extends Base
         try {
             Object idFromValues = values.get("id");
             if (idFromValues == null) {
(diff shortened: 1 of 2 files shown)


shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
```diff
@@ -463,7 +463,7 @@ public S updatePartial(@RequestBody Map<String, Object> values, @PathVariable("i
         try {
             Object idFromValues = values.get("id");
             if (idFromValues == null) {
-                LOG.error("Field 'id' is missing in the passed values.", entityId, values);
+                LOG.error("Field 'id' (entity {})is missing in the passed values: {}.", entityId, values);
```
dnlkoch, line 466:
```suggestion
                LOG.error("Field 'id' is missing in the passed values: {}", values);
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
```diff
@@ -508,7 +507,7 @@ public S updatePartial(@RequestBody Map<String, Object> values, @PathVariable("i
                     ade
             );
         } catch (NumberFormatException nfe) {
-            LOG.error("Can't parse 'id' field ({}) from values ({}). It has to be an Integer.",
+            LOG.error("Can't parse 'id' field ({}) from values ({}). It has to be an Integer.: {}",
                 values, entityId, nfe.getMessage());
```
dnlkoch, line 511:
```suggestion
            LOG.error("Can't parse 'id' field from passed values ({}). It has to be an integer: {}",
                values, nfe.getMessage());
```

dnlkoch (approved):

shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
```diff
@@ -472,10 +472,9 @@ public S updatePartial(@RequestBody Map<String, Object> values, @PathVariable("i
                 throw new ResponseStatusException(HttpStatus.BAD_REQUEST);
             }
 
-            Optional<S> persistedEntity = service.findOne(entityId);
-
-            if (persistedEntity.isPresent()) {
-                S updatedEntity = service.updatePartial(entityId, values);
+            S persistedEntity = service.findOne(entityId).orElseThrow();
+            if (persistedEntity != null) {
+                S updatedEntity = service.updatePartial(entityId, persistedEntity, values);
```
KaiVolland, line 477: The update method includes the same check. Do we have to update this too?

ahennr: In `updatePartial` is checked whether the `entityId` matches the id of the persisted entity. Beside, the entity has to be passed to `updatePartial` otherwise the used target type in `hasPermission` is `Map<String, Object>` but it must be `S` (a sub-class of `BaseEntity`)

PR #262, opened 2021-01-29
Partly fix mvn site and friends

marcjansen:
This PR suggests …

* …to add a new property `maven-site-plugin.version` for accessing the version of the maven site plugin
* …to additionally add the site plugin to the parent pom under `plugins`

These changes make `mvn site` work (currently broken on master). When the `reporting` profile is active (`mvn -P reporting test site`, e.g.), several information from the project info lugin will be included (That was previousl already included).

Nonetheless several things aren't work as expected win the genrated report, these should be tackled some other day.

Please review.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,11 +65,12 @@
 
     <!-- Plugins -->
     <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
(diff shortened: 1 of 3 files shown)


dnlkoch (approved):
Thanks @marcjansen!

marcjansen:
Thanks for the review, @dnlkoch

PR #266, opened 2021-02-04
JacksonMapper: Extend config from ObjectMapper to initialize objectMapper used for Hibernate and GraphQL correctly

ahennr:
During startup of a SHOGun boot application, two `objectMapper` instances are created currently: The one for hibernate and the one of Spring (used in GraphQL).
However, since `hibernate.types.jackson.object.mapper` is set in `hibernate.properties` the `objectMapper` created in JacksonConfig is used *before* the one of Spring is created. Therefore, the passed values for `shogun.srid` and `shogun.coordinatePrecisionScale` are not yet available which leads to a erroneous configured mapper for geometries (SRID:0).
By this PR, `JacksonConfig` extends `ObjectMapper` and uses its methods directly and fixes the broken initialization procedure. In addition, `readValue` needed to be overwritten in order to guarantee that search for `JsonSuperType` annotation already has been done.

Plz review @terrestris/devs

### DIFF
--- a/shogun-boot/pom.xml
+++ b/shogun-boot/pom.xml
@@ -119,6 +119,11 @@
       <groupId>org.keycloak</groupId>
       <artifactId>keycloak-spring-boot-starter</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>postgresql</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/shogun-boot/src/test/java/JacksonConfigTest.java
+++ b/shogun-boot/src/test/java/JacksonConfigTest.java
@@ -1,0 +1,59 @@
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import de.terrestris.shogun.boot.config.ApplicationConfig;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.locationtech.jts.geom.LineString;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+
+@SpringBootTest(classes = ApplicationConfig.class, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("test")
+public class JacksonConfigTest {
+
+    @Value("${shogun.srid}")
+    private int srid;
+
+    @Autowired
+    ObjectMapper objectMapper;
+
+    @Test
+    public void assertThatObjectMapperHasJtsModuleRegistered() {
+        var registeredModules = objectMapper.getRegisteredModuleIds();
+        boolean hasJtsModuleRegistered = false;
+        for (Object module : registeredModules) {
+            if (StringUtils.equalsIgnoreCase((CharSequence) module, "com.bedatadriven.jackson.datatype.jts.JtsModule")) {
+                hasJtsModuleRegistered = true;
+            }
+        }
+        Assertions.assertTrue(hasJtsModuleRegistered, "JTS Module is not registrered in ObjectMapper");
+    }
+
+    @Test
+    public void assertThatObjectMapperHasCorrectlyConfiguredJtsModule() throws JsonProcessingException {
+        String LINESTRING_25832 = "{" +
+            "\"type\": \"LineString\"," +
+            "\"coordinates\": [" +
+                "[" +
+                    "711957.369742162," +
+                    "5637657.304058334" +
+                "]," +
+                "[" +
+                    "711979.3599907478," +
+                    "5637629.050147795" +
+                "]," +
+                "[" +
+                    "712000.5099907471," +
+                    "5637596.860147787" +
+                "]" +
+            "]" +
+        "}";
+        LineString parsedLineString = objectMapper.readValue(LINESTRING_25832, LineString.class);
+        Assertions.assertEquals(parsedLineString.getSRID(), srid, "Coordinate reference system of parsed " +
+            "geometry does not match the one configured in JTS Module");
+    }
+
+}
--- a/shogun-boot/src/test/java/de/terrestris/shogun/boot/config/JacksonConfigTest.java
+++ b/shogun-boot/src/test/java/de/terrestris/shogun/boot/config/JacksonConfigTest.java
@@ -1,22 +1,32 @@
+package de.terrestris.shogun.boot.config;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import de.terrestris.shogun.boot.config.ApplicationConfig;
+import de.terrestris.shogun.boot.runner.ApplicationInitializer;
 import org.apache.commons.lang3.StringUtils;
+import org.junit.AfterClass;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.locationtech.jts.geom.LineString;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.context.ActiveProfiles;
 
-@SpringBootTest(classes = ApplicationConfig.class, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@SpringBootTest(classes = { ApplicationConfig.class, JdbcConfiguration.class }, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @ActiveProfiles("test")
 public class JacksonConfigTest {
+
+    // Replace ApplicationInitializer by functionless mock
+    @MockBean
+    private ApplicationInitializer applicationInitializer;
 
     @Value("${shogun.srid}")
     private int srid;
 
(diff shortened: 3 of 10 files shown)


KaiVolland (approved):
Nice. Just one minor – none blocking – note.

simonseyock:
This looks great, but is it possible to provide an `ObjectMapper` instead of an `ObjectMapperSupplier`? The docs are very thin in that regard: https://vladmihalcea.com/hibernate-types-customize-jackson-objectmapper/

As you mentioned the problem lies in the fact that hibernate asks for the objectmapper before spring is initialized. For the JsonB interfaces it is important the modified object mapper is used. The config values are something I did not think about enough.

How are you assuring that the objectmapper gets created before spring is intialized but also gets the correct config values?

simonseyock:
To test if this does work correctly, you have to try to deserialize data from the database to a shogun model. Like having an `UserDetails` like this:

```java
@Data
@ToString
@EqualsAndHashCode
    private String text;
}
```

and then having an `User` entity in the database with `"{"text":"test"}"` as value for `user_details`.

I found this issue that might be interesting for this issue: https://github.com/hibernate/hibernate-orm/pull/2649

simonseyock:
Would it be possible to create the object mapper for hibernate early, save it as a static variable, and then initialize the modules and subtypes later when spring is intialized?

simonseyock:
My propsal would be to revert back to the previous implementation which implemented `ObjectSupplier` but there not intializing the objectmapper in the `get` method, but just return the static objectmapper and do the initialization in an init method that is annotated with `@PostConstruct`. This way we could also be sure that everything is completed on startup of the application instead of doing it on the first read. The reflection takes some time and that is why it would be nice if that was done in the startup.

hwbllmnn:
> My propsal would be to revert back to the previous implementation which implemented `ObjectSupplier` but there not intializing the objectmapper in the `get` method, but just return the static objectmapper and do the initialization in an init method that is annotated with `@PostConstruct`. This way we could also be sure that everything is completed on startup of the application instead of doing it on the first read. The reflection takes some time and that is why it would be nice if that was done in the startup.

I think one of the problems is that the object mapper is used by Hibernate _during startup_, so that's not really an option. Also, the `@PostConstruct` is not executed for the Hibernate object mapper, since that one is constructed via reflection IIRC.

simonseyock:
From my observations hibernate requests multiple `ObjectMapper`s during startup but they are not used for entity deserializations yet. The timeline should look like this:
* Hibernate requests an `ObjectMapper` from the `ObjectMapperSupplier`. This happens multiple times.
* Spring starts up and requests an `ObjectMapper` Bean.
* The deserialization happens on requests.

I was thinking to use the previous approach but combining that with the late initialization that @ahennr introduced, like this:
```java
class JacksonConfig extends ObjectMapperSupplier {
  private static ObjectMapper _objectMapper;
  
  @Override
  public ObjectMapper get() {
    if (_objectMapper == null) {
      _objectMapper = new ObjectMapper():
    }
    return _objectMapper;
  }
  
  @PostConstruct
  public void init() {
    // init modules here
  }
}
```

Additionally we would need to add the appropriate annotations for Spring (`@Component` ?) and expose the static `ObjectMapper` also as a `@Bean`.

I think it is important to make sure that the initialization is only done once so the costly reflections are only executed once.

ahennr:
> Would it be possible to create the object mapper for hibernate early, save it as a static variable, and then initialize the modules and subtypes later when spring is intialized?

Unfortunately this lead to the same problem: The hibernate serializer is created having a `GeometryFactory` `with SRID:0` which is wrong .

hwbllmnn:
At least in the project we've been looking at some serialization already happened during startup (i.e. before any requests). Personally I don't think that having the initialization code run multiple times isn't really a problem, it actually runs pretty fast and also just happens (as far as we saw twice) on startup.

ahennr:
> I think it is important to make sure that the initialization is only done once so the costly reflections are only executed once.

Therefore variable `initialized` was introduced which is set  to `true` after first successful initialization. In following calls, `init` is omitted.

simonseyock:
> Unfortunately this lead to the same problem: The hibernate serializer is created having a GeometryFactory with SRID:0 which is not correct at this point.
I was thinking about adding the GeometryFactory only during `@PostConstruct` there the spring config should be set or am I wrong? Maybe we would not change them to be not static.

simonseyock:
For me it boils down to the question if hibernate can really work here with a raw `ObjectMapper` or if it actually needs the `ObjectMapperSupplier`. This was the only thing I could find about this in the docs: https://vladmihalcea.com/hibernate-types-customize-jackson-objectmapper/

simonseyock:
> At least in the project we've been looking at some serialization already happened during startup (i.e. before any requests). Personally I don't think that having the initialization code run multiple times isn't really a problem, it actually runs pretty fast and also just happens (as far as we saw twice) on startup.

The class path search for `@JsonSuperType` actually takes some time.

hwbllmnn:
> > Unfortunately this lead to the same problem: The hibernate serializer is created having a GeometryFactory with SRID:0 which is not correct at this point.
> 
> I was thinking about adding the GeometryFactory only during `@PostConstruct` there the spring config should be set or am I wrong? Maybe we would need the config variables to change them to be not static.

That's actually one problem, the `@PostConstruct` is not executed for the hibernate object mapper instance.

simonseyock:
To extend my example:

```java
@Component
class JacksonConfig extends ObjectMapperSupplier {
  private static ObjectMapper _objectMapper;
  
  @Value("${shogun.srid:4326}")
  protected int srid;

  @Value("${shogun.coordinatePrecisionScale:10}")
  protected int coordinatePrecisionScale;
  
  @Override
  public ObjectMapper get() {
    if (_objectMapper == null) {
      _objectMapper = new ObjectMapper():
    }
    return _objectMapper;
  }
  
  @Bean
  public ObjectMapper objectMapper() {
    return get();
  }
  
  @PostConstruct
  public void init() {
    _objectMapper.registerModule(jtsModule());

    _objectMapper.configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false);
    _objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
    _objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
    _objectMapper.setDateFormat(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX"));

    for (var entry : findAnnotatedClasses().entrySet()) {
        _objectMapper.addMixIn(entry.getKey(), entry.getValue());
    }
  }
}
```

simonseyock:
> 
> 
> > > Unfortunately this lead to the same problem: The hibernate serializer is created having a GeometryFactory with SRID:0 which is not correct at this point.
> > 
> > 
> > I was thinking about adding the GeometryFactory only during `@PostConstruct` there the spring config should be set or am I wrong? Maybe we would need the config variables to change them to be not static.
> 
> That's actually one problem, the `@PostConstruct` is not executed for the hibernate object mapper instance.

That's why I want to have only one static instance :D

hwbllmnn:
> > > > Unfortunately this lead to the same problem: The hibernate serializer is created having a GeometryFactory with SRID:0 which is not correct at this point.
> > > 
> > > 
> > > I was thinking about adding the GeometryFactory only during `@PostConstruct` there the spring config should be set or am I wrong? Maybe we would need the config variables to change them to be not static.
> > 
> > 
> > That's actually one problem, the `@PostConstruct` is not executed for the hibernate object mapper instance.
> 
> That's why I want to have only one static instance :D

Since the creation of the hibernate object mapper happens first (when the Spring initialization has not been finished yet), that would be the broken one ;-)

simonseyock:
> Since the creation of the hibernate object mapper happens first (when the Spring initialization has not been finished yet), that would be the broken one ;-)

But the jtsModule doesn't need to be added on creation. It can be added later.

simonseyock:
If you look in the example this should happen:
* hibernate requests an objectmapper. The static variable is intialized
* spring intializes. In `@PostConstruct` the module and other configuration is applied to the static objectmapper.
* Spring requests the objectmapper it is supplied the same static one.

simonseyock:
Please note that in my example I do not extend `ObjectMapper` but implement `ObjectMapperSupplier`.

hwbllmnn:
> Please note that in my example I do not extend `ObjectMapper` but implement `ObjectMapperSupplier`.

That sounds good in theory, but unfortunately the `ObjectMapperSupplier` interface isn't actually used AFAICT. As far as we debugged last week, there will always be two `ObjectMapper` instances for the moment. In the comments on the blog post you mentioned there is actually a discussion about this, and as mentioned that will only be possible if [this PR](https://github.com/hibernate/hibernate-orm/pull/3400) is merged.

simonseyock:
Hmm, when I was debugging that, the `ObjectMapperSupplier` was used. For the project I tested it with it was actually deserializing the objects correctly from the database. If you want we could look into that together at some point.

ahennr:
After a discussion with @simonseyock and @hwbllmnn , usage of `ObjectMapperSupplier` interface has been reintroduced c617ccd and initialization of `ObjectMapper` to an initialization method (in particular registration of modules, etc.) that is called some time after construction of initial mapper.

Plz review @simonseyock  @hwbllmnn

simonseyock (approved):

dnlkoch:
Changes are fine to me, but I would love to see a regression test for the correct initialization of the object mapper.

ahennr:
> Changes are fine to me, but I would love to see a regression test for the correct initialization of the object mapper.

Alright, I'll have a look at this as soon as possible.

ahennr:
@dnlkoch : In c6bf92b and e3b2d70 I've added a `@SpringBootTest` that starts a simple SHOGun boot application (using testcontainers) in order to check that JtsModule has been configured properly during application setup by parsing a simple GeoJSON using the enhanced `objectMapper` instance. Does this match your requested issue?

dnlkoch (approved):
This is awesome @ahennr!

ahennr:
Thx for your reviews guys. 👍🏼

PR #306, opened 2021-04-16
Introduce getGroupsForUser() returning groups of user in SecurityContext 

ahennr:
This PR introduces a new method `getGroupsForUser()` to the `SecurityContextUtil` that returns the groups of the user that is actually logged in. In contrast to the existing one, it uses the `otherClaims` map of the `IdToken` / `AccessToken` in `KeycloakPrincipal` (Keycloak plugin https://github.com/terrestris/keycloak-oidc-group-uuid-mapper required and activated for the client in specific realm).
As fallback, the already existing method for the user if claims are not provided.

Plz review @terrestris/devs


ahennr:
BTW: Tested with keycloak 12.0.4 in `shogun-docker`. Maybe https://github.com/terrestris/shogun/pull/284 can be merged afterwards as well?

dnlkoch:
Agreed, #284 is merged.

PR #307, opened 2021-04-16
Cleanup parent POM

dnlkoch:
This updates the parent POM by:

* Fixing a typo in variable `jib-maven-plugin.version`, was `jib-maven-pluign.version`.
* Placing all versions in variables.
* Downgrading the explicit `spring-boot` dependency from `2.4.0` to `2.3.5` to match the version of the parent POM.
* Removing all properties that where already defined in the parent `spring-boot-dependencies` POM.
* Removing all dependencies that where already defined in the parent `spring-boot-dependencies` POM.

The latter ensures we always get the tested/supported versions by the current spring-boot version. Besides the spring-boot version downgrade (see above), no noteable version up- or downgrades were detected during cleanup.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,31 +77,21 @@
     <java.version>11</java.version>
 
     <!-- Plugins -->
-    <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
-    <maven-deploy-plugin.version>3.0.0-M1</maven-deploy-plugin.version>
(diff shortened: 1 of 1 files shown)


hwbllmnn (approved):

ahennr (approved):

LukasLohoff (approved):

dnlkoch:
Thanks for the reviews!

PR #312, opened 2021-05-03
Allows saving of files to filesystem

KaiVolland:
This introduces interfaces and methods to upload files to the filesystem instead of adding them as bytearray to the DB.

- Includes a migration to add the `path` property to the `File` entity.
- Includes config options to the `UploadProperties`
  - `path` --> The basePath of the upload directory
  - `maxSize` --> The maximum size of upload (currently not used)

## BREAKING CHANGE:

- Removes `validateFile` from the `FileUtil` and moved validation to the `BaseFileService`

### DIFF
--- a/shogun-config/src/main/java/de/terrestris/shogun/properties/UploadProperties.java
+++ b/shogun-config/src/main/java/de/terrestris/shogun/properties/UploadProperties.java
@@ -20,7 +20,6 @@ import lombok.Data;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.stereotype.Component;
 
 @Data
 @Configuration
@@ -32,5 +31,9 @@ public class UploadProperties {
 
     @NestedConfigurationProperty
     private ImageFileUploadProperties image;
+
+    private String path;
+
+    private String maxSize;
 
 }
--- a/shogun-config/src/main/java/de/terrestris/shogun/properties/UploadProperties.java
+++ b/shogun-config/src/main/java/de/terrestris/shogun/properties/UploadProperties.java
@@ -32,7 +32,7 @@ public class UploadProperties {
     @NestedConfigurationProperty
     private ImageFileUploadProperties image;
 
-    private String path;
+    private String basePath;
 
     private String maxSize;
 
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
@@ -123,6 +123,8 @@ public abstract class BaseFileController<T extends BaseFileService<?, S>, S exte
                     LOG.trace("… load file from disk");
                     byte[] fileByteArray = FileUtils.readFileToByteArray(dataFile);
                     return new ResponseEntity<>(fileByteArray, responseHeaders, HttpStatus.OK);
+                } else {
+                    LOG.error("Could not load File {} from disk", file.getId());
                 }
             }
 
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
@@ -18,8 +18,6 @@ package de.terrestris.shogun.lib.controller;
 
 import de.terrestris.shogun.lib.model.File;
 import de.terrestris.shogun.lib.service.BaseFileService;
-import de.terrestris.shogun.lib.util.FileUtil;
-import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -41,7 +39,7 @@ public abstract class BaseFileController<T extends BaseFileService<?, S>, S exte
 
     protected final Logger LOG = LogManager.getLogger(getClass());
 
-    @Value("${upload.path}")
+    @Value("${upload.basePath}")
     protected String uploadBasePath;
 
     @Autowired
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
@@ -107,21 +107,10 @@ public abstract class BaseFileController<T extends BaseFileService<?, S>, S exte
                 responseHeaders.setContentType(MediaType.parseMediaType(file.getFileType()));
                 responseHeaders.setContentDisposition(ContentDisposition.parse(
                     String.format("inline; filename=\"%s\"", file.getFileName())));
-
                 LOG.trace("Successfully got file with UUID {}", fileUuid);
-                if (file.getPath() == null) {
-                    LOG.trace("… load file from database");
-                    return new ResponseEntity<>(file.getFile(), responseHeaders, HttpStatus.OK);
-                }
 
-                java.io.File dataFile = new java.io.File(uploadBasePath + "/" + file.getPath());
-                if (dataFile.exists()) {
-                    LOG.trace("… load file from disk");
-                    byte[] fileByteArray = FileUtils.readFileToByteArray(dataFile);
-                    return new ResponseEntity<>(fileByteArray, responseHeaders, HttpStatus.OK);
-                } else {
-                    LOG.error("Could not load File {} from disk", file.getId());
-                }
+                byte[] fileData = service.getFileData(file);
+                return new ResponseEntity<>(fileData, responseHeaders, HttpStatus.OK);
             }
 
             LOG.error("Could not find entity of type {} with UUID {}", getGenericClassName(), fileUuid);
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
@@ -212,7 +212,7 @@ public abstract class BaseFileController<T extends BaseFileService<?, S>, S exte
     @PostMapping(value = "/uploadToFileSystem", consumes = "multipart/form-data")
     @ResponseStatus(HttpStatus.CREATED)
     public S addToFileSystem(MultipartFile uploadedFile) {
-        LOG.debug("Requested to upload a multipart-file");
+        LOG.debug("Requested to upload a multipart-file and to save it to the file system");
 
         try {
 
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
@@ -33,13 +33,9 @@ import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
 import org.springframework.web.server.ResponseStatusException;
 
-import java.io.ByteArrayInputStream;
-import java.io.InputStream;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
-
-import static de.terrestris.shogun.lib.util.FileUtil.convertToFile;
 
 public abstract class BaseFileController<T extends BaseFileService<?, S>, S extends File> {
 
@@ -175,7 +171,7 @@ public abstract class BaseFileController<T extends BaseFileService<?, S>, S exte
 
         try {
 
-            service.isValidType(uploadedFile.getContentType());
+            service.isValid(uploadedFile);
 
             S persistedFile = service.create(uploadedFile);
 
@@ -218,7 +214,7 @@ public abstract class BaseFileController<T extends BaseFileService<?, S>, S exte
 
         try {
 
-            service.isValidType(uploadedFile.getContentType());
+            service.isValid(uploadedFile);
 
             S persistedFile = service.create(uploadedFile, true);
 
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
@@ -18,9 +18,12 @@ package de.terrestris.shogun.lib.controller;
 
 import de.terrestris.shogun.lib.model.File;
 import de.terrestris.shogun.lib.service.BaseFileService;
+import de.terrestris.shogun.lib.util.FileUtil;
+import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.MessageSource;
 import org.springframework.context.i18n.LocaleContextHolder;
 import org.springframework.core.GenericTypeResolver;
@@ -30,13 +33,20 @@ import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
 import org.springframework.web.server.ResponseStatusException;
 
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 
+import static de.terrestris.shogun.lib.util.FileUtil.convertToFile;
+
 public abstract class BaseFileController<T extends BaseFileService<?, S>, S extends File> {
 
     protected final Logger LOG = LogManager.getLogger(getClass());
+
+    @Value("${upload.path}")
+    protected String uploadBasePath;
 
     @Autowired
     protected T service;
@@ -97,31 +107,36 @@ public abstract class BaseFileController<T extends BaseFileService<?, S>, S exte
 
             if (entity.isPresent()) {
                 S file = entity.get();
-
-                LOG.info("Successfully got file with UUID {}", fileUuid);
-
                 final HttpHeaders responseHeaders = new HttpHeaders();
                 responseHeaders.setContentType(MediaType.parseMediaType(file.getFileType()));
                 responseHeaders.setContentDisposition(ContentDisposition.parse(
                     String.format("inline; filename=\"%s\"", file.getFileName())));
 
-                return new ResponseEntity<>(file.getFile(), responseHeaders, HttpStatus.OK);
-            } else {
-                LOG.error("Could not find entity of type {} with UUID {}",
-                    getGenericClassName(), fileUuid);
+                LOG.trace("Successfully got file with UUID {}", fileUuid);
+                if (file.getPath() == null) {
+                    LOG.trace("… load file from database");
+                    return new ResponseEntity<>(file.getFile(), responseHeaders, HttpStatus.OK);
+                }
 
-                throw new ResponseStatusException(
-                    HttpStatus.NOT_FOUND,
-                    messageSource.getMessage(
-                        "BaseController.NOT_FOUND",
-                        null,
-                        LocaleContextHolder.getLocale()
-                    )
-                );
+                java.io.File dataFile = new java.io.File(uploadBasePath + "/" + file.getPath());
+                if (dataFile.exists()) {
+                    LOG.trace("… load file from disk");
+                    byte[] fileByteArray = FileUtils.readFileToByteArray(dataFile);
+                    return new ResponseEntity<>(fileByteArray, responseHeaders, HttpStatus.OK);
+                }
             }
+
+            LOG.error("Could not find entity of type {} with UUID {}", getGenericClassName(), fileUuid);
+            throw new ResponseStatusException(
+                HttpStatus.NOT_FOUND,
+                messageSource.getMessage(
+                    "BaseController.NOT_FOUND",
+                    null,
+                    LocaleContextHolder.getLocale()
+                )
+            );
         } catch (AccessDeniedException ade) {
-            LOG.info("Access to entity of type {} with UUID {} is denied",
-                getGenericClassName(), fileUuid);
+            LOG.info("Access to entity of type {} with UUID {} is denied", getGenericClassName(), fileUuid);
 
             throw new ResponseStatusException(
                 HttpStatus.NOT_FOUND,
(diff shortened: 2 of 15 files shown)


shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,63 @@ public File create(MultipartFile uploadFile) throws Exception {
...
+        byte[] fileByteArray = FileUtil.fileToByteArray(uploadFile);
+        java.io.File outFile = new java.io.File(fileDirectory, filename);
+        InputStream in = new ByteArrayInputStream(fileByteArray);
+
+        try (OutputStream out = new FileOutputStream(outFile)) {
+            IOUtils.copy(in, out);
+            LOG.info("Saved file with id {} to {}: ", savedFile.getId(), savedFile.getPath());
+        } catch (Exception e) {
+            LOG.error("Error when saving file {} to disk: " + e.getMessage(), savedFile.getId());
+            throw e;
```
KaiVolland, line 110: Do we have to manually rollback the creation here?

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,63 @@ public File create(MultipartFile uploadFile) throws Exception {
...
+            return this.create(uploadFile);
+        }
+
+        String uploadBasePath = uploadProperties.getPath();
+        String filename = uploadFile.getOriginalFilename();
+
+        FileUtil.validateFile(uploadFile);
+        File file = new File();
+        file.setFileType(uploadFile.getContentType());
+        file.setFileName(filename);
```
LukasLohoff, line 88:
```suggestion
        file.setFileName(fileName);
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,63 @@ public File create(MultipartFile uploadFile) throws Exception {
...
+        file.setActive(true);
+
+        File savedFile = this.create(file);
+
+        UUID fileUuid = savedFile.getFileUuid();
+
+        // Setup path and directory
+        String path = fileUuid + "/" + filename;
```
LukasLohoff, line 96:
```suggestion
        String path = fileUuid + "/" + fileName;
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,63 @@ public File create(MultipartFile uploadFile) throws Exception {
...
+        java.io.File fileDirectory = new java.io.File(uploadBasePath + "/" + fileUuid);
+        fileDirectory.mkdirs();
+
+        // Write multipart file data to target directory
+        byte[] fileByteArray = FileUtil.fileToByteArray(uploadFile);
+        java.io.File outFile = new java.io.File(fileDirectory, filename);
```
LukasLohoff, line 102:
```suggestion
        java.io.File outFile = new java.io.File(fileDirectory, fileName);
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,63 @@ public File create(MultipartFile uploadFile) throws Exception {
...
+     * @return
+     * @throws Exception
+     */
+    public File create(MultipartFile uploadFile, Boolean writeToSystem) throws Exception {
+        if (!writeToSystem) {
+            return this.create(uploadFile);
+        }
+
+        String uploadBasePath = uploadProperties.getPath();
+        String filename = uploadFile.getOriginalFilename();
```
LukasLohoff, line 83: - I think we should validate that `uploadBasePath` and `filename` aren't empty here (unless that isn't done already at another time)
- do we need to check / sanitize `fileName` if we use the filename as part of the final file path? thinking of e.g. special characters 
    - maybe it's easier and safer to not use the filename at all - if we don't really need it

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,63 @@ public File create(MultipartFile uploadFile) throws Exception {
         return savedFile;
     }
 
+    /**
+     * Creates a new File entity in the Database.
+     * The file is stored as file on the .
```
LukasLohoff, line 70:
```suggestion
     * The file is stored as file on the disk.
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
LukasLohoff: either that, or we have to create and save the entity after the file was saved to disk

but in that case we could be left with a file without an entity in case `this.create()` fails...

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,60 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
+            LOG.info("Saved file with id {} to {}: ", savedFile.getId(), savedFile.getPath());
+        } catch (Exception e) {
+            LOG.error("Error when saving file {} to disk: " + e.getMessage(), savedFile.getId());
+            throw e;
+        }
+
+        // Update entity with saved File
+        savedFile.setPath(path);
+        this.repository.save(savedFile);
+
```
LukasLohoff, line 150:
```suggestion
```
(double newline)

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,60 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
     }
 
+    @Override
+    public ImageFile create(MultipartFile uploadFile, Boolean writeToSystem) throws Exception {
+        if (!writeToSystem) {
+            return this.create(uploadFile);
+        }
+
+        String uploadBasePath = uploadProperties.getPath();
+        String filename = uploadFile.getOriginalFilename();
```
LukasLohoff, line 105: see comments to `FileService`

LukasLohoff (commented):
Looks great! Few questions on validation of fileName, see comments

marcjansen:
Can we include file content validation (is this actually an image...) here or is this sth. for another day?

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,60 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
+            IOUtils.copy(in, out);
+            LOG.info("Saved file with id {} to {}: ", savedFile.getId(), savedFile.getPath());
+        } catch (Exception e) {
+            LOG.error("Error when saving file {} to disk: " + e.getMessage(), savedFile.getId());
+            throw e;
+        }
+
+        // Update entity with saved File
+        savedFile.setPath(path);
+        this.repository.save(savedFile);
```
LukasLohoff, line 149: I think it's better to return the output of `this.repository.save(savedFile)` instead of `return savedFile;`

KaiVolland:
> Can we include file content validation (is this actually an image...) here or is this sth. for another day?

We do content**Type** validation. I don't know if content validation adds more security/functionality and if its even doable with an acceptable effort.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
KaiVolland: As we use the  uuid for the file location i don't think we can't upload the file and then create it. But the rollback should be a oneline...

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -80,41 +80,47 @@ public File create(MultipartFile uploadFile, Boolean writeToSystem) throws Excep
...
 
         String uploadBasePath = uploadProperties.getPath();
-        String filename = uploadFile.getOriginalFilename();
+        if(uploadBasePath != null) {
+            throw new Exception("Could not upload file. uploadBasePath is null.");
+        }
+        String fileName = uploadFile.getOriginalFilename();
+        if(fileName != null) {
+            throw new Exception("Could not upload file. fileName is null.");
+        }
```
marcjansen, line 89: Are we sure the conditionals are correct?

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -102,12 +102,16 @@ public ImageFile create(MultipartFile uploadFile, Boolean writeToSystem) throws
...
 
         String uploadBasePath = uploadProperties.getPath();
-        String filename = uploadFile.getOriginalFilename();
+        if(uploadBasePath != null) {
+            throw new Exception("Could not upload file. uploadBasePath is null.");
+        }
+        String fileName = uploadFile.getOriginalFilename();
+        if(fileName != null) {
+            throw new Exception("Could not upload file. fileName is null.");
+        }
```
marcjansen, line 111: Same here. Either the conditional or the message is wrong

marcjansen:
Thanks for e55e17e, I do not know `Tika`, but this looks solid. It would be good to test this with a variety of different uploads / browsers, because `!mediaType.toString().equals(contentType)` *might* give false positives.

The easiest thing would surely be to have this being configurable, whether this validation should happen. And probably also being easily able to plug in a common project validation. This all might e out of scope if this issue.

Please proceed as you like, either including the (configurable) validation or by letting it out. My 2 ¢

shogun-config/src/main/java/de/terrestris/shogun/properties/UploadProperties.java
```diff
@@ -33,4 +32,8 @@
     @NestedConfigurationProperty
     private ImageFileUploadProperties image;
 
+    private String path;
```
ahennr, line 35: as `path` is a reserved keyword in PostgreSQL, I would prefer to rename this variable.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,69 @@ public File create(MultipartFile uploadFile) throws Exception {
...
+     * @return
+     * @throws Exception
+     */
+    public File create(MultipartFile uploadFile, Boolean writeToSystem) throws Exception {
+        if (!writeToSystem) {
+            return this.create(uploadFile);
+        }
+
+        String uploadBasePath = uploadProperties.getPath();
+        if(uploadBasePath == null || uploadBasePath.isEmpty()) {
```
ahennr, line 83:
```suggestion
        if(StringUtils.isEmpty(uploadBasePath)) {
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/File.java
```diff
@@ -67,4 +67,8 @@
     @Column(length = Integer.MAX_VALUE)
     @Getter @Setter
     private byte[] file;
+
+    @Column
+    @Getter @Setter
+    private String path;
```
ahennr, line 73: Should be path of the file be ignored in JSON?

shogun-config/src/main/java/de/terrestris/shogun/properties/UploadProperties.java
marcjansen: nice, Röö, goiod catch, even though it would have been escaped properly, this can be the source of oddities. Cool.

marcjansen:
Another general question: Will this scale well? I still remember that filesystems and containerized systems aren't the best friends. Is the behavior configurable, where the actual file ends up.

shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
```diff
@@ -97,31 +107,36 @@
...
+                    LOG.trace("… load file from database");
+                    return new ResponseEntity<>(file.getFile(), responseHeaders, HttpStatus.OK);
+                }
+
+                java.io.File dataFile = new java.io.File(uploadBasePath + "/" + file.getPath());
+                if (dataFile.exists()) {
+                    LOG.trace("… load file from disk");
+                    byte[] fileByteArray = FileUtils.readFileToByteArray(dataFile);
+                    return new ResponseEntity<>(fileByteArray, responseHeaders, HttpStatus.OK);
+                }
```
dnlkoch, line 126: Please add a logger if the file could not be found.

shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
```diff
@@ -97,31 +107,36 @@
...
             }
+
```
dnlkoch, line 128: I would also prefer to move the above code to the service, but this is not blocking.

shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
```diff
@@ -194,6 +209,49 @@ public S add(MultipartFile uploadedFile) {
         }
     }
 
+    @PostMapping(value = "/uploadToFileSystem", consumes = "multipart/form-data")
+    @ResponseStatus(HttpStatus.CREATED)
+    public S addToFileSystem(MultipartFile uploadedFile) {
+        LOG.debug("Requested to upload a multipart-file");
```
dnlkoch, line 215:
```suggestion
        LOG.debug("Requested to upload a multipart-file and to save it to the file system");
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/File.java
dnlkoch: I agree, this is an internal information and shouldn't be needed externally.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseFileService.java
```diff
@@ -20,15 +20,17 @@
...
 
 import java.util.Optional;
 import java.util.UUID;
 
 public abstract class BaseFileService<T extends BaseFileRepository<S, Long> & JpaSpecificationExecutor<S>, S extends File> extends BaseService<T, S> implements IBaseFileService<T, S> {
 
     @PostAuthorize("hasRole('ROLE_ADMIN') or hasPermission(returnObject.orElse(null), 'READ')")
-    public Optional<S> findOne(UUID fileUuid) {
-        return repository.findByFileUuid(fileUuid);
+    public Optional<S> findOne(UUID fileUuid) { return repository.findByFileUuid(fileUuid);
```
dnlkoch, line 31: Please undo.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
```diff
@@ -51,6 +65,69 @@ public File create(MultipartFile uploadFile) throws Exception {
...
+            throw new Exception("Could not upload file. uploadBasePath is null.");
+        }
+        String fileName = uploadFile.getOriginalFilename();
+        if(fileName == null || fileName.isEmpty()) {
```
dnlkoch, line 87:
```suggestion
        if (StringUtils.isEmpty(fileName)) {
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,63 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
+        String uploadBasePath = uploadProperties.getPath();
+        if(uploadBasePath == null || uploadBasePath.isEmpty()) {
+            throw new Exception("Could not upload file. uploadBasePath is null.");
+        }
+        String fileName = uploadFile.getOriginalFilename();
+        if(fileName == null || fileName.isEmpty()) {
+            throw new Exception("Could not upload file. fileName is null.");
+        }
+
+        FileUtil.validateFile(uploadFile);
```
dnlkoch, line 113: Can we move the validation to the `isValidType` method which is running on start?

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,63 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
+
+        // Write multipart file data to target directory
+        java.io.File outFile = new java.io.File(fileDirectory, fileName);
+        InputStream in = new ByteArrayInputStream(fileByteArray);
+
+        try (OutputStream out = new FileOutputStream(outFile)) {
+            IOUtils.copy(in, out);
+            LOG.info("Saved file with id {} to {}: ", savedFile.getId(), savedFile.getPath());
+        } catch (Exception e) {
+            LOG.error("Error when saving file {} to disk: " + e.getMessage(), savedFile.getId());
```
dnlkoch, line 147:
```suggestion
            LOG.error("Error while saving file {} to disk: {}", e.getMessage(), savedFile.getId());
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,63 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
+        java.io.File fileDirectory = new java.io.File(uploadBasePath + "/" + fileUuid);
+        fileDirectory.mkdirs();
+
+        // Write multipart file data to target directory
+        java.io.File outFile = new java.io.File(fileDirectory, fileName);
+        InputStream in = new ByteArrayInputStream(fileByteArray);
+
+        try (OutputStream out = new FileOutputStream(outFile)) {
+            IOUtils.copy(in, out);
+            LOG.info("Saved file with id {} to {}: ", savedFile.getId(), savedFile.getPath());
```
dnlkoch, line 145:
```suggestion
            LOG.info("Saved file with id {} to: {}", savedFile.getId(), savedFile.getPath());
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,63 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
+        } catch (Exception e) {
+            LOG.error("Error when saving file {} to disk: " + e.getMessage(), savedFile.getId());
+            LOG.info("Rollback creation of file {}.", savedFile.getId());
+            this.repository.delete(savedFile);
```
dnlkoch, line 149: We should also remove the `fileDirectory`.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -89,4 +95,63 @@ public void isValidType(String contentType) throws InvalidContentTypeException {
...
+            throw e;
+        }
+
+        // Update entity with saved File
+        savedFile.setPath(path);
+        return this.repository.save(savedFile);
```
dnlkoch, line 155: Setting the path can also be done before calling `this.create(file)`.

KaiVolland: The path is composed by the uuid and the fileName. I think the uuid is generate by the `create` function.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/FileService.java
LukasLohoff: I think most characters are allowed in linux-based systems. but what happens if you modify the request and change `fileName` to something that contains slashes?

- will that create a subfolder?
- will the file creation fail?
- can you somehow break out of the current directory?

shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseFileController.java
KaiVolland: 0f92b05

LukasLohoff (approved):
:+1:

ahennr (approved):

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -64,29 +72,77 @@ public ImageFile create(MultipartFile uploadFile) throws Exception {
...
+        String uploadBasePath = uploadProperties.getBasePath();
+        if(StringUtils.isEmpty(uploadBasePath)) {
+            throw new Exception("Could not upload file. uploadBasePath is null.");
         }
-
-        if (uploadProperties.getImage().getSupportedContentTypes() == null) {
-            throw new InvalidContentTypeException("No list of supported content types for the image file upload found. " +
-                "Please check your application.yml");
+        String fileName = uploadFile.getOriginalFilename();
+        if(StringUtils.isEmpty(fileName)) {
```
dnlkoch, line 86:
```suggestion
        if (StringUtils.isEmpty(fileName)) {
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
```diff
@@ -64,29 +72,77 @@ public ImageFile create(MultipartFile uploadFile) throws Exception {
...
+    public ImageFile create(MultipartFile uploadFile, Boolean writeToSystem) throws Exception {
+        if (!writeToSystem) {
+            return this.create(uploadFile);
         }
 
-        if (uploadProperties.getImage() == null) {
-            throw new InvalidContentTypeException("No properties for the image file upload found. " +
-                "Please check your application.yml");
+        String uploadBasePath = uploadProperties.getBasePath();
+        if(StringUtils.isEmpty(uploadBasePath)) {
```
dnlkoch, line 82:
```suggestion
        if (StringUtils.isEmpty(uploadBasePath)) {
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/ImageFileService.java
dnlkoch: Right, ignore the comment.

shogun-lib/src/test/java/de/terrestris/shogun/lib/service/BaseFileServiceTest.java
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    public void isValidFileName_shouldAllowRegularFilenames() {
+        try {
+            fileService.isValidFileName("Peter.pdf");
+            fileService.isValidFileName("Peter.png");
+            fileService.isValidFileName("Mein Spezial Bild mit Leerzeichen.png");
+            fileService.isValidFileName("Mein_Spezial_Bild_mit_Unterstrichen.png");
+            fileService.isValidFileName("Mein-Spezial-Bild-mit-Bindestrichen.png");
+            fileService.isValidFileName("@klammeraffe#hastag.png");
+            fileService.isValidFileName("@klammeraffe#hastag.png");
```
dnlkoch, line 56:
```suggestion
            fileService.isValidFileName("lord-fauntleroy.png");
```

dnlkoch (approved):

PR #356, opened 2021-07-15
Adjust several JSONB typings

dnlkoch:
:rotating_light: **BREAKING CHANGE** :rotating_light:

This updates some JSONB typings in the `Application` and `Layer` classes:

* Rename interface `LayerToolConfig` to `ApplicationToolConfig`.
* Make both fields `toolConfig` and `layerConfig` capable of storing lists of `ApplicationToolConfig` and `LayerConfig`.
* Introduce `GeoJsonObject` types for storing `features` in the `Layer` class.


shogun-lib/src/main/java/de/terrestris/shogun/lib/model/Layer.java
```diff
@@ -73,7 +61,7 @@
     @Column(columnDefinition = "jsonb")
     @Basic(fetch = FetchType.LAZY)
     @ToString.Exclude
-    private List<LayerFeature> features;
+    private GeoJsonObject features;
```
simonseyock, line 64: I suppose `GeoJsonObject` can hold arbitrary feature related data in a dictionary or similar. Can it hold typed data as well?

simonseyock (approved):
One question which could also addressed at some later point, apart from that it looks good to me.

annarieger (approved):

shogun-lib/src/main/java/de/terrestris/shogun/lib/model/Layer.java
dnlkoch: Actually it can hold an object shaped like a GeoJSON (see [here](https://github.com/opendatalab-de/geojson-jackson/blob/master/src/main/java/org/geojson/GeoJsonObject.java)).

simonseyock: I phrased this a little bad I think. The GeoJSON objects can normally hold a properties object. I was asking myself if these are planned to use and if these should get typed. After looking at the code I saw that this is not possible. This is something we will loose, but I think it still nicer to just a standard here.

dnlkoch: Yes, you're correct, the `properties` aren't typed in the `Feature` class, but I think this should be fine for most scenarios.

dnlkoch:
Thanks for the reviews!

PR #389, opened 2021-10-04
introduce some java 12-17 features & minor refactorings

LukasLohoff:
Contains some minor refactorings:

- uses multiline blocks
- uses switch expressions
- replaces `logManager` with lombok's `@Log4j2` annotation (:exclamation: possibly breaking change)
- fixes startup script: specifies mvn output and changes address back due to redirect


ahennr:
🔝

PR #390, opened 2021-10-04
Fix workdir

hwbllmnn:
Fix the dev location of shogun-boot and upgrade packages on build.

Please review @terrestris/devs

PR #393, opened 2021-10-05
Fix dev dockerfile

dnlkoch:
This suggests:

* To make use of the same openjdk base image as in production (referenced by the jib plugin). 
* To set the correct workdir path and
* To set the default application servlet path to `/`.

Replaces #390 and requires [#21](https://github.com/terrestris/shogun-docker/pull/21) (shogun-docker).

### DIFF
--- a/.github/workflows/startup.yml
+++ b/.github/workflows/startup.yml
@@ -14,6 +14,9 @@ jobs:
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


hwbllmnn (approved):

LukasLohoff:
Startup action will be repaired in another MR

PR #398, opened 2021-10-06
use fixed shogun docker version

LukasLohoff:
- uses fixed shogun-docker version
   - this ensures reproducibility
   - version should be updated regularly
- small improvements to wait.sh script


.github/workflows/startup.yml
```diff
@@ -34,8 +34,10 @@ jobs:
         run: mvn install --log-file mvn.out.txt
 
       - name: Clone shogun-docker
-        run: git clone https://github.com/terrestris/shogun-docker
-        working-directory: ../
+        uses: actions/checkout@v2
+        with:
+          repository: terrestris/shogun-docker
+          ref: 4fdcbc51fa2d20bcfb0bcafafdfdece83c863819
```
ahennr, line 40: Will this ref be updated automatically / dependency bot?

LukasLohoff: I'll try to configure it like that

dnlkoch (approved):

PR #463, opened 2022-02-07
Configurable authentication and identity provider

dnlkoch:
## Description

Currently SHOGun is capable to deal with Keycloak as authentication provider only. Even if Keycloak supports a broad range of external connectors (LDAP, OAuth2, …) in some scenarios it is required to become compatible with a custom provider.

This MR suggests to make all existing Keycloak related operations optional.

### Disable/Enable built-in Keycloak support

To decide whether to make use of Keycloak or not, a new property `keycloak.enabled` (default is to `true`) has been introduced. Example configuration for the `application.yml`:

```
keycloak:
  enabled: false
```

In case of `true` default implementations of the Keycloak integration in SHOGun will be used. For `false` projects need to implement several methods by itself (see ahead).

### Provider services

Two new interfaces have been introduced that hold the methods required in the context of authentication:

- `GroupProviderService`
   - Default implementation: `KeycloakGroupProviderService`
- `UserProviderService`
   - Default implementation: `KeycloakUserProviderService`

#### BootWebSecurityConfig

Previously SHOGun applications needed to extend the `BootWebSecurityConfig`, now it's either:

- `KeycloakBootWebSecurityConfig` if Keycloak is active
- `SimpleBootWebSecurityConfig` if Keycloak is not active

#### SecurityContextUtil

All Keycloak specific methods have been moved to the `UserProviderService` or `KeycloakUtil`:

- `getUserBySession()` -> `UserProviderService.getUserBySession()`
- `getUserFromAuthentication()` -> `UserProviderService.getUserFromAuthentication()`
- `getKeycloakUserIdFromAuthentication()` -> `KeycloakUtil.getKeycloakUserIdFromAuthentication()`
- `getGroupsForUser -> `GroupProviderService.findByUser`
- `isUserInGroup()` -> removed

#### Services

- Removed deprecated method `GroupService.findByKeycloakId()`
- `GroupService.setTransientKeycloakRepresentations()` -> `GroupProviderService.setTransientRepresentations()`
- `UserService.findOrCreateByKeyCloakId()` -> `UserProviderService.findOrCreateByProviderId()`

#### User

- The field `keycloakId` is renamed to `authProviderId`
- The field `keycloakRepresentation` is renamed to `providerDetails`
- The `User` class requires a type argument for the authentication provider details object.

#### Group

- The field `keycloakId` is renamed to `authProviderId`
- The field `keycloakRepresentation` is renamed to `providerDetails`
- The `Group` class requires a type argument for the authentication provider details object.

#### GroupRepository

- `findByKeycloakId()` has been renamed to `findByAuthProviderId()`

#### UserRepository

- `findByKeycloakId()` has been renamed to `findByAuthProviderId()`

#### GroupController

- GET endpoint `/groups/keycloak/{id}` has been removed

--

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist

### DIFF
--- a/shogun-boot/src/main/java/de/terrestris/shogun/boot/config/KeycloakBootWebSecurityConfig.java
+++ b/shogun-boot/src/main/java/de/terrestris/shogun/boot/config/KeycloakBootWebSecurityConfig.java
@@ -1,0 +1,66 @@
+/* SHOGun, https://terrestris.github.io/shogun/
+ *
+ * Copyright © 2020-present terrestris GmbH & Co. KG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package de.terrestris.shogun.boot.config;
+
+import de.terrestris.shogun.config.DefaultWebSecurityConfig;
+import org.keycloak.adapters.springsecurity.KeycloakConfiguration;
+import org.keycloak.adapters.springsecurity.authentication.KeycloakAuthenticationProvider;
+import org.keycloak.adapters.springsecurity.config.KeycloakWebSecurityConfigurerAdapter;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.core.authority.mapping.SimpleAuthorityMapper;
+import org.springframework.security.core.session.SessionRegistryImpl;
+import org.springframework.security.web.authentication.session.RegisterSessionAuthenticationStrategy;
+import org.springframework.security.web.authentication.session.SessionAuthenticationStrategy;
+
+@ConditionalOnExpression("${keycloak.enabled:true}")
+@Configuration
+@EnableWebSecurity
+@KeycloakConfiguration
+public class KeycloakBootWebSecurityConfig extends KeycloakWebSecurityConfigurerAdapter implements DefaultWebSecurityConfig {
+
+    @Autowired
+    public void configureGlobal(AuthenticationManagerBuilder auth) {
+        SimpleAuthorityMapper grantedAuthorityMapper = new SimpleAuthorityMapper();
+        grantedAuthorityMapper.setPrefix("ROLE_");
+
+        grantedAuthorityMapper.setConvertToUpperCase(true);
+
+        KeycloakAuthenticationProvider keycloakAuthenticationProvider = keycloakAuthenticationProvider();
+        keycloakAuthenticationProvider.setGrantedAuthoritiesMapper(grantedAuthorityMapper);
+        auth.authenticationProvider(keycloakAuthenticationProvider);
+    }
+
+    @Bean
+    @Override
+    protected SessionAuthenticationStrategy sessionAuthenticationStrategy() {
+        return new RegisterSessionAuthenticationStrategy(new SessionRegistryImpl());
+    }
+
+    @Override
+    public void configure(HttpSecurity http) throws Exception {
+        super.configure(http);
+
+        customHttpConfiguration(http);
+    }
+
+}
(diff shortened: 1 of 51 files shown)


shogun-boot/src/main/java/de/terrestris/shogun/boot/config/KeycloakBootWebSecurityConfig.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.security.core.authority.mapping.SimpleAuthorityMapper;
+import org.springframework.security.core.session.SessionRegistryImpl;
+import org.springframework.security.web.authentication.session.RegisterSessionAuthenticationStrategy;
+import org.springframework.security.web.authentication.session.SessionAuthenticationStrategy;
+
+@ConditionalOnExpression("${keycloak.enabled:true}")
+@Configuration
+@EnableWebSecurity
+@KeycloakConfiguration
+public class KeycloakBootWebSecurityConfig extends KeycloakWebSecurityConfigurerAdapter implements DefaultWebSecurityConfig {
```
LukasLohoff, line 39: Did you see that the keycloak spring adapter will be deprecated soon? See

- https://www.keycloak.org/2022/02/adapter-deprecation
- https://www.keycloak.org/docs/latest/release_notes/index.html#spring-security-and-boot-adapter-deprecation
- https://www.springcloud.io/post/2022-02/keycloak-announces/#gsc.tab=0
- https://www.baeldung.com/spring-security-openid-connect

Just a general remark, and probably something for another MR. But maybe this will make the rework easier? Or more complicated? I'm not entirely sure about the impact on this...

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/GroupService.java
```diff
@@ -66,90 +59,46 @@ public List<Group> findAllBy(Specification specification) {
...
-
     /**
      * Finds a Group by the passed keycloak ID. If it does not exists in the SHOGun DB it gets created.
      *
      * @param keycloakGroupId
      * @return
      */
     @Transactional
-//    @PreAuthorize("hasRole('ROLE_ADMIN') or hasPermission(#keycloakGroupId, 'CREATE')")
+    @PreAuthorize("hasRole('ROLE_ADMIN') or hasPermission(#keycloakGroupId, 'CREATE')")
```
LukasLohoff, line 93: if you're wondering why this was commented: I think there was an issue when shogun was out-of-sync with keycloak i.e. when a group existed in keycloak but not in shogun db

if that happens while a normal user is authenticated certain requests may fail because that user doesn't have create permission for `keycloakGroupId`.

LukasLohoff: maybe it's enough to check if the authenticated user can read the group?

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/provider/keycloak/KeycloakUserProviderService.java
```diff
@@ -0,0 +1,165 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.List;
+import java.util.Optional;
+
+import static de.terrestris.shogun.lib.util.KeycloakUtil.getKeycloakUserIdFromAuthentication;
+
+/**
+ * NOTE: Make sure not to use services here, else the security checks will not run on them due to circular
```
LukasLohoff, line 47: OT: I just remebered I wrote a test for this. Will open a PR soon

LukasLohoff (approved):
Looks nice! I added some general comments / questions

shogun-boot/src/main/java/de/terrestris/shogun/boot/config/KeycloakBootWebSecurityConfig.java
dnlkoch: Good point, haven't noticed that during refactoring. Actually I'm not sure if this would become a trivial update, so I suggest to handle it in an upcoming PR, but I'll check it anyway.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/provider/keycloak/KeycloakUserProviderService.java
dnlkoch: Nice!

shogun-boot/src/main/java/de/terrestris/shogun/boot/config/KeycloakBootWebSecurityConfig.java
LukasLohoff: :+1:

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/GroupService.java
LukasLohoff: I think the GroupService / UserService may be invoked by keycloak events: https://github.com/terrestris/shogun/blob/main/shogun-lib/src/main/java/de/terrestris/shogun/lib/listener/KeycloakEventListener.java#L38

In this case there will be no authenticated user

LukasLohoff: Nevermind, I just looked at the new classed again and saw that this is still the same for `KeycloakUserProviderService`

annarieger (approved):

dnlkoch:
Thanks for the reviews @annarieger @LukasLohoff!

PR #522, opened 2022-05-25
feat: allow cacheService to evict specific regions

LukasLohoff:
## Description

Allows cacheController / cacheService to evict specific regions. i.e.:

```bash
curl -k -v -X POST 'https://localhost/cache/evict?regions=layers,groups,files'
```

`CacheService::evictCacheRegions` can be called with multiple regions:

```java
@Autowired
CacheService cacheService;

cacheService.evictCacheRegions("layers", "topics", "groups");
```

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Tests
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/CacheController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/CacheController.java
@@ -51,7 +51,7 @@ public class CacheController {
 
         try {
             if (regions == null || regions.isEmpty()) {
-                service.evictAllCacheRegions();
+                service.evictCache();
                 log.info("Successfully evicted all cache regions.");
             } else {
                 service.evictCacheRegions(regions.toArray(new String[]{}));
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/CacheController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/CacheController.java
@@ -61,7 +61,7 @@ public class CacheController {
             return ResponseEntity.ok().build();
         } catch (Exception e) {
             log.error("Could not evict the cache: {}", e.getMessage());
-            log.trace("Full stack trace:", e);
+            log.trace("Full stack trace: ", e);
 
             throw new ResponseStatusException(
                 HttpStatus.INTERNAL_SERVER_ERROR,
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/CacheController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/CacheController.java
@@ -26,8 +26,11 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.server.ResponseStatusException;
+
+import java.util.List;
 
 @Log4j2
 @RestController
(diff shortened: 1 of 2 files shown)


shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/CacheController.java
```diff
@@ -42,19 +45,23 @@ public class CacheController {
...
+            } else {
+                service.evictCacheRegions(regions.toArray(new String[]{}));
+                log.info("Successfully evicted cache regions {}", regions);
+            }
 
             return ResponseEntity.ok().build();
         } catch (Exception e) {
             log.error("Could not evict the cache: {}", e.getMessage());
-            log.trace("Full stack trace: {}", e);
+            log.trace("Full stack trace:", e);
```
dnlkoch, line 64:
```suggestion
            log.trace("Full stack trace: ", e);
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/CacheService.java
```diff
@@ -43,4 +47,23 @@ public void evictCache() throws Exception {
...
         cache.evictAllRegions();
     }
+
+    public void evictCacheRegions(String... region) throws Exception {
+        if (region == null) {
+            return;
+        }
+        if (entityManager == null) {
+            throw new Exception("Could not get the entity manager.");
+        } else {
```
dnlkoch, line 57: `else` is not necessarily needed.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/CacheService.java
```diff
@@ -43,4 +47,23 @@ public void evictCache() throws Exception {
...
+            EntityManagerFactory entityManagerFactory = entityManager.getEntityManagerFactory();
+            SessionFactory sessionFactory = entityManagerFactory.unwrap(SessionFactory.class);
+            Cache cache = sessionFactory.getCache();
+            for (String r : region) {
+                if (StringUtils.isEmpty(r)) {
+                    continue;
+                }
+                cache.evictRegion(r);
```
dnlkoch, line 65: What happens for an unavailable region identifier? Do we need to check if the region exists before?

dnlkoch (approved):
This is a very useful addition, thanks @LukasLohoff!

I added just some minor remarks and a question regarding the (required?) check for the region existence. If not needed, just ignore.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/CacheService.java
LukasLohoff: indeed! removed in e2e58ba45085a61b7588768bf83814757a683e13

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/CacheService.java
LukasLohoff: good point. I'll do some testing

LukasLohoff: A NullPointerException is thrown if one region identifier is invalid, preventing the following regions from being cleared. AFAIK there is no hibernate method which checks if an identifier is valid or not.

I could wrap `cache.evictRegion(r);` into a try-catch block. That way, other (possible) valid regions will still be cleared.

LukasLohoff: implemented here: 9cfa96669971e5da3e06bdba9a5a1270dd3bc7b2

dnlkoch (approved):

PR #549, opened 2022-07-26
create shogun users in `getUserBySession` if needed

LukasLohoff:
## Description

Adds a fallback in `getUserBySession` which creates a shogun user if it does not exist yet.

Usually, the user should have been created from an event fired when the user was created in keycloak. But I think we can't always rely on that event; e.g.:

- user was created via import
- event listener plugin was not active

Also changes the generic `@RequestMapping("/")` to `@GetMapping` in ResourceController . This should be more secure. I don't think we need to allow other request types there.

@terrestris/devs Please review.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Tests
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/shogun-boot/src/main/java/de/terrestris/shogun/boot/controller/ResourceController.java
+++ b/shogun-boot/src/main/java/de/terrestris/shogun/boot/controller/ResourceController.java
@@ -35,7 +35,7 @@ public class ResourceController {
     @Autowired(required = false)
     BuildProperties buildProperties;
 
-    @RequestMapping("/")
+    @GetMapping("/")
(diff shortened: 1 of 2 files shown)


simonseyock (approved):
The call to `getKeycloakUserIdFromAuthentication` ensures that this will return an empty optional if no user is logged in, so I think this is good to merge.

dnlkoch (approved):

LukasLohoff:
Thanks for the reviews!

PR #606, opened 2022-11-15
breaking: add paging and improve permission check

LukasLohoff:
## Description

- this PR implements pagination for the `findAll` method in `BaseController` and `BaseFileController`
- it also contains a refactoring of how the permissions are checked
   -  previously, the `@PostFilter` annotation meant that all entities had to be retrieved from the database, and afterwards a permission check had to be performed for each entity. This was very slow for a large number of entities
   - the new method performs the following checks to determine if the permission check can be skipped:
      - does the user have admin rights?
      - does the user have permissions through class permissions (either group oder user)?
   - if not, the actual permission check is now integrated into the spring data JPA queries to prevent fetching more entities than needed

## TODOs:

- [x] check group class permissions before querying entities
- [x] check group instance permissions when querying entities
- [x] write migration guide
- [x] check swagger API
- [x] add tests

Anything graphql related will be adressed in another MR: https://github.com/terrestris/shogun/pull/662

### 1. `BaseController::findAll()` - method signature has changed

- the method now returns `Page<BaseEntity>` instead of `List<BaseEntity>`
- this means the results are now wrapped into a spring `Page` object, e.g.:
   ```json
  {
    "content": [
      {
        "id": 475870,
        "created": "2022-10-07T14:01:43.11027Z"
        [...]
      }
    ],
    "pageable": {
      "sort": {
        "empty": true,
        "sorted": false,
        "unsorted": true
      },
      "offset": 0,
      "pageNumber": 0,
      "pageSize": 1,
      "paged": true,
      "unpaged": false
    },
    "last": false,
    "totalElements": 2,
    "totalPages": 2,
    "size": 1,
    "number": 0,
    "sort": {
      [...]
    },
    "first": true,
    "numberOfElements": 1,
    "empty": false
  }
   ```
   - `content` contains the list of entities
   - the object also contains metadata, for more information see https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Page.html
- the method has a new optional parameter `Pageable pageable` which can be used to specify which page and size is returned e.g.:
   ```json
  {
    "page": 0,
    "size": 10
  }
   ```
- if no pagination configuration is provided, one page with all entities is returned. But it is recommended to use pagination for performance reasons

### 2. Custom permission evaluators have to implement a `findAll` method

- if your project uses custom permission evaluators, you have to implement this method to provide a way to check permissions for requests with pagination
- BaseEntityPermissionEvaluator contains a default implemenation which performs the new improved permisison check described above
   - the default only works for the "shogun way" (permission managment through UserInstance-, UserClass- GroupInstance- and GroupClassPermissions)

Follow-up which also adds pagination for GraphQL-Endpoints and introduces cursor based pagination (WIP): https://github.com/terrestris/shogun/pull/662

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Tests
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist

### DIFF
--- a/shogun-boot/src/main/java/de/terrestris/shogun/boot/config/ApplicationConfig.java
+++ b/shogun-boot/src/main/java/de/terrestris/shogun/boot/config/ApplicationConfig.java
@@ -27,7 +27,6 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
-import org.springframework.security.data.repository.query.SecurityEvaluationContextExtension;
 
 @SpringBootApplication
 @EnableJpaRepositories(
@@ -42,11 +41,6 @@ import org.springframework.security.data.repository.query.SecurityEvaluationCont
     ImageFileUploadProperties.class
 })
 public class ApplicationConfig {
-
-    @Bean
-    public SecurityEvaluationContextExtension securityEvaluationContextExtension() {
-        return new SecurityEvaluationContextExtension();
-    }
 
     public static void main(String[] args) {
         SpringApplication.run(ApplicationConfig.class, args);


shogun-boot/src/main/java/de/terrestris/shogun/boot/config/BootKeycloakConfig.java
```diff
@@ -18,8 +18,10 @@
 
 import de.terrestris.shogun.config.KeycloakConfig;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.context.annotation.Bean;
```
dnlkoch, line 21:
```suggestion
```

shogun-boot/src/main/java/de/terrestris/shogun/boot/config/BootKeycloakConfig.java
```diff
@@ -18,8 +18,10 @@
...
 import org.springframework.context.annotation.Configuration;
+import org.springframework.security.data.repository.query.SecurityEvaluationContextExtension;
```
dnlkoch, line 23:
```suggestion
```

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/provider/keycloak/KeycloakUserProviderService.java
```diff
@@ -47,7 +47,7 @@
  */
 @ConditionalOnExpression("${keycloak.enabled:true}")
 @Log4j2
-@Component
+@Component("keycloakUserProviderService")
```
dnlkoch, line 50: Is this really needed (since it should be the default)?

shogun-lib/src/main/java/de/terrestris/shogun/lib/security/access/entity/BaseEntityPermissionEvaluator.java
```diff
@@ -230,4 +242,80 @@ public boolean hasPermissionByGroupClassPermission(User user, BaseEntity entity,
...
+     * @return A page of entities.
+     */
+    @Override
+    public Page<E> findAll(User user, Pageable pageable, BaseCrudRepository<E, Long> repository, Class<E> baseEntityClass) {
+        // option A: user has role `ADMIN`
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+
+        List<GrantedAuthority> authorities = new ArrayList<>(authentication.getAuthorities());
+        var isAdmin = authorities.stream().anyMatch(
+            grantedAuthority -> StringUtils.endsWithIgnoreCase(grantedAuthority.getAuthority(), "ADMIN")
```
dnlkoch, line 265: An `endsWith` check might result in false positives, I would suggest to check against the full role name with `equalsIgnoreCase`.

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
```diff
@@ -72,18 +76,41 @@
...
+        // note: security check is done in permission evaluator
+        Optional<User> userOpt = userProviderService.getUserBySession();
+
+        // todo: can this be simplified? autowiring BaseEntityPermissionEvaluator did not work.
+        BaseEntityPermissionEvaluator entityPermissionEvaluator =
+            this.getPermissionEvaluatorForClass(BaseEntity.class.getCanonicalName());
+
+        Class<? extends BaseEntity> entityClass = this.getBaseEntityClass();
+
+        return entityPermissionEvaluator.findAll(userOpt.orElseThrow(), pageable, repository, entityClass);
```
dnlkoch, line 99: Do we really want to throw if no user is available? Project implementations might allow anonymous access e.g. to the `/applications` interface where this could fail. Maybe it's safer to provide the check inside the evaluator (which might needed to be implemented in such scenario anyway)?

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
```diff
@@ -72,18 +76,41 @@
...
     public List<S> findAll() {
         return (List<S>) repository.findAll();
     }
 
+    @Transactional(readOnly = true)
+    public Page<S> findAll(Pageable pageable) {
+        // note: security check is done in permission evaluator
+        Optional<User> userOpt = userProviderService.getUserBySession();
+
+        // todo: can this be simplified? autowiring BaseEntityPermissionEvaluator did not work.
```
dnlkoch, line 93: Actually I don't have any other idea…

dnlkoch (commented):
Thanks @LukasLohoff! This looks awesome so far!

shogun-lib/src/main/java/de/terrestris/shogun/lib/security/access/entity/BaseEntityPermissionEvaluator.java
LukasLohoff: fixed in 63741cd4c7a03a279856fdbdf9e73df0cd855a95

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/security/provider/keycloak/KeycloakUserProviderService.java
LukasLohoff: is not needed, removed in fixed in 63741cd4c7a03a279856fdbdf9e73df0cd855a95

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
LukasLohoff: moved the check into the default implementation in fixed in 63741cd4c7a03a279856fdbdf9e73df0cd855a95

shogun-lib/src/main/java/de/terrestris/shogun/lib/service/BaseService.java
LukasLohoff: ok, it's fine for me

LukasLohoff:
Thanks for the review @dnlkoch
Addressed your suggestions in 63741cd4c7a03a279856fdbdf9e73df0cd855a95 and 63741cd4c7a03a279856fdbdf9e73df0cd855a95

dnlkoch (approved):

PR #608, opened 2022-11-17
breaking: fix serialization issues in BaseService#update and use `JsonPatch` for `updatePartial`

### LINKED ISSUE #450: `BaseService.update` returns object that is not a hibernate proxy
The `BaseService.update` method serializes the given entity, deserializes it again and saves it. This causes the resulting object to not be a proper hibernate proxy. (edit: to be more precise: any related objects are no hibernate proxies anymore, but merely the objects that got deserialized. It can be that the object only contains an Id if the field was using the `BaseEntityIdSerializer`).   Also all other managed instances of the object are broken as well.

Firstly the question arises why this code with the serialization and deserialization exists:
```java
    @PreAuthorize("hasRole('ROLE_ADMIN') or hasPermission(#entity, 'UPDATE')")
    @Transactional(isolation = Isolation.SERIALIZABLE)
    public S update(Long id, S entity) throws IOException {
        Optional<S> persistedEntityOpt = repository.findById(id);

        ObjectNode jsonObject = objectMapper.valueToTree(entity);

        // Ensure the created timestamp will not be overridden.
        S persistedEntity = persistedEntityOpt.orElseThrow();
        OffsetDateTime createdTimestamp = persistedEntity.getCreated();
        String serialized = createdTimestamp != null ? createdTimestamp.toInstant().toString() : null;
        jsonObject.put("created", serialized);

        S updatedEntity = objectMapper.readerForUpdating(persistedEntity).readValue(jsonObject);

        return repository.save(updatedEntity);
    }
```

I suppose this in place to leave the entity intact. Is this really neccessary?

The simplest way to achieve the same would be:
```java
  entity.setCreated(persistedEntity.getCreated());
```

Another way to solve this would be to use the lombok `@SuperBuilder` annotation and create a copy. See https://www.projectlombok.org/features/experimental/SuperBuilder

For our project we used the following workaround:
```java
            var updatedEntity = super.update(entityId, entity);

            entityManager.flush(); // flush database changes
            entityManager.detach(updatedEntity); // this way the broken object is no longer a managed hibernate object
            // without the detach the findById would just return the broken managed object
            updatedEntity = repository.findById(entityId).orElseThrow(); // get a new managed hibernate proxy object
```

This might not be feasible for the `BaseService` as I assume that `flush` does have performance issues if called often.

LukasLohoff:
## Description

Fixes issues with properties being overridden by removing the serialization in `BaseService#update`.

Extracted from #515:

>  Herein the `update` method is updated to not serialize the update candidate anymore. In addition the `updatePartial` has been reworked to make use of [`JSON Patch`](https://datatracker.ietf.org/doc/html/rfc7386) and the [`@DynamicUpdate`](https://www.baeldung.com/spring-data-jpa-dynamicupdate) annotation is presented for all entities.

### 1. `BaseService#updatePartial` changed its method signature.

Old:

```java
public S updatePartial(Long entityId, S entity, Map<String, Object> values)
```

New:

```java
public S updatePartial(S entity, JsonMergePatch patch)
```

Projects which override `updatePartial` or use it directly need to update their code accordingly.

### 2. `BaseController#updatePartial` changed its signature

Projects using the endpoint `PATCH /{entityClass}/{id}` should update their client code:

- supplying the entity id in the json body is no longer necessary

### 3. Annotate your hibernate models with `@DynamicUpdate`

For more information see https://www.baeldung.com/spring-rest-json-patch#http-patch-request

- from #515 
- fixes #450 
   -  `updatePartial` still uses serialization and

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Tests
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [] No

## Checklist

### DIFF
--- a/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
+++ b/shogun-lib/src/main/java/de/terrestris/shogun/lib/controller/BaseController.java
@@ -16,15 +16,11 @@
  */
 package de.terrestris.shogun.lib.controller;
 
+import com.github.fge.jsonpatch.mergepatch.JsonMergePatch;
 import de.terrestris.shogun.lib.controller.security.permission.BasePermissionController;
 import de.terrestris.shogun.lib.model.BaseEntity;
 import de.terrestris.shogun.lib.service.BaseService;
 import lombok.extern.log4j.Log4j2;
-
-import java.time.OffsetDateTime;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.MessageSource;
 import org.springframework.context.i18n.LocaleContextHolder;
@@ -37,6 +33,11 @@ import org.springframework.security.access.AccessDeniedException;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.server.ResponseStatusException;
 
+import java.time.OffsetDateTime;
+import java.util.List;
+import java.util.Optional;
+
+// TODO Specify and type extension of BaseService
 @Log4j2
 public abstract class BaseController<T extends BaseService<?, S>, S extends BaseEntity> extends BasePermissionController<T, S> {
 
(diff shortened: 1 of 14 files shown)


LukasLohoff:

dnlkoch (approved):

PR #694, opened 2023-05-22
feat: add search field config to control used search engines

annarieger:
## Description

Introduce `SearchFieldConfig` field to `DefaultApplicationClientConfig` jsonb model to control, which search engines should be used with an application's search field. Currently the following two options are enabled:
* `useNominatim` -> whether to use the OSM search engine for the address data
* `useSolr` -> whether to use the Apache Solr engine for the layer data search

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Tests
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

- [ ] I understand and agree that the changes in this PR will be licensed under the 
- [ ] I have followed the [guidelines for contributing](https://github.com/terrestris/shogun/blob/main/CONTRIBUTING.md).
- [ ] The proposed change fits to the content of the [code of conduct](https://github.com/terrestris/shogun/blob/main/CODE_OF_CONDUCT.md).


shogun-lib/src/main/java/de/terrestris/shogun/lib/model/jsonb/application/DefaultApplicationClientConfig.java
```diff
@@ -59,4 +59,9 @@ public class DefaultApplicationClientConfig implements ApplicationClientConfig {
     )
     private String defaultLanguage;
 
+    @Schema(
+        description = "The configuration of the search field."
+    )
+    private DefaultSearchFieldConfig searchFieldConfig;
```
dnlkoch, line 65: I think this might become confusing as there is already the [DefaultApplicationToolConfig](https://github.com/terrestris/shogun/blob/main/shogun-lib/src/main/java/de/terrestris/shogun/lib/model/jsonb/application/DefaultApplicationToolConfig.java) for tool configurations available. Is there any reason why we don't make use of it?

PR #886, opened 2024-06-27
fix: use containeraddress for jwk-set-uri

FritzHoing:
## Description

Needs to be updated in shogun-docker also.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Tests
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

- [x] I have added a screenshot/screencast to illustrate the visual output of my update.

### DIFF
--- a/shogun-config/src/main/resources/application-base.yml
+++ b/shogun-config/src/main/resources/application-base.yml
@@ -91,7 +91,7 @@ spring:
       resourceserver:
         jwt:
           issuer-uri: https://${KEYCLOAK_HOST:shogun-keycloak}/auth/realms/SHOGun
-          jwk-set-uri: https://${KEYCLOAK_HOST:shogun-keycloak}/auth/realms/SHOGun/protocol/openid-connect/certs
+          jwk-set-uri: http://shogun-keycloak:8080/auth/realms/SHOGun/protocol/openid-connect/certs
   graphql:
     schema:
       locations: classpath*:graphql/**/
@@ -112,7 +112,7 @@ keycloak:
   client-id: shogun-boot
   principal-attribute: preferred_username
   disableHostnameVerification: true
-  internal-server-url: shogun-keycloak
+  internal-server-url: "shogun-keycloak:8080"
   extract-roles-from-resource: true
   extract-roles-from-realm: false
 


simonseyock:
This problem occurs if the `KEYCLOAK_HOST` is not reachable from inside the container.
The problem started to appear with the newest SHOGun version.
I noticed the same while using the helm-charts and opened this PR there: https://github.com/terrestris/helm-charts/pull/90

shogun-config/src/main/resources/application-base.yml
```diff
@@ -91,7 +91,7 @@ spring:
       resourceserver:
         jwt:
           issuer-uri: https://${KEYCLOAK_HOST:shogun-keycloak}/auth/realms/SHOGun
-          jwk-set-uri: https://${KEYCLOAK_HOST:shogun-keycloak}/auth/realms/SHOGun/protocol/openid-connect/certs
+          jwk-set-uri: http://shogun-keycloak:8080/auth/realms/SHOGun/protocol/openid-connect/certs
```
dnlkoch, line 94: Not sure why this is needed, but removing the environment variable `KEYCLOAK_HOST` entirely seems not to be the right way to me.

dnlkoch (changes requested):

shogun-config/src/main/resources/application-base.yml
simonseyock: The problem is that spring is accessing the `jwk-set-uri` to autoconfigure itself and retrieve some certificates from keycloak. If the `KEYCLOAK_HOST` is used, this only works if the `KEYCLOAK_HOST` is accessible from inside the container.

Using the container address here seems to be the solution, as this should be working in any environment.

The documentation is not exhaustive in that regard, but I understand it this way:
* The `issuer-url` is used to check the `issuer` field in the jwt. If `jwk-set-uri` is not set, this is also used to autoconfigure and retrieve the certificates.
* The `jwk-set-uri` is used instead for the requests if set

https://docs.spring.io/spring-security/reference/reactive/oauth2/resource-server/jwt.html#webflux-oauth2resourceserver-jwt-jwkseturi

PR #988, opened 2025-02-25
Include keycloak-js assets

dnlkoch:
## Description

Keycloak versions >= 26 [don't include the keycloak-js assets](https://www.keycloak.org/docs/latest/upgrading/index.html#keycloak-js) anymore. This suggests to include the assets in the shogun package.

Needed as soon as [#225](https://github.com/terrestris/shogun-docker/pull/225) is available.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [x] Dependency updates
- [ ] Tests
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/shogun-boot/src/main/resources/public/assets/lib/keycloak/keycloak-authz.d.ts
+++ b/shogun-boot/src/main/resources/public/assets/lib/keycloak/keycloak-authz.d.ts
@@ -1,0 +1,143 @@
+/*
+ * MIT License
+ *
+ * Copyright 2017 Brett Epps <https://github.com/eppsilon>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the
+ * following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or substantial
+ * portions of the Software.
+ *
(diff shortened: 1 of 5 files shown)


KaiVolland (approved):
Is there an npm package or somehting we could use here? Mainting/Updating this might be annoying when the assets need updates for new keycloak versions.

dnlkoch:
Yes, it is [available](https://www.npmjs.com/package/keycloak-js) and I struggled while including the assets directly as well. But since the index page is just an ordinary HTML page I'm also somewhat uncertain about creating a npm package for it. This would e.g. require `npm` in the build lifecycle. 

Do you have a idea on how to deal with that?

KaiVolland:
It should be possible to just download the [`dist.tarball`](https://registry.npmjs.org/keycloak-js/26.1.3) of the package. In this case it would be https://registry.npmjs.org/keycloak-js/-/keycloak-js-26.1.3.tgz.

Then extract the needed files. Either by reading the package.json or directly using the `/package/lib/`. 

We could put this into a shell script. Not sure if this is better then including npm, but it shouldn't be to difficult.

dnlkoch:
I'll check if this is possible without further changes in the build pipeline.

KaiVolland (approved):
Nice. But i think you have to add the license header yourself. :)

dnlkoch:
Updated the code to actually download the keycloak js adapter (and all other assets required by the landing page) during the project initialization phase (in particular during `mvn compile`).

KaiVolland (approved):

hwbllmnn (approved):

